PR #4, opened 2020-10-21
DOC: add estimating pi example to usage

hagenw:
Would be nice to have an usage example inside the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,6 +13,16 @@ The package is lightweight,
 and has the small tqdm_ package
 as it's only external dependency.
 
+Code example,
+that lists all WAV files in the ``data`` folder:
+
+.. code-block:: python
+
+    import audeer
+
+    files = audeer.list_file_names('data', filetype='wav')
+
+
 .. _tqdm: https://tqdm.github.io/
 
 .. badges images and links:
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -1,10 +1,38 @@
 Usage
 =====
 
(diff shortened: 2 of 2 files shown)


hagenw:
If you look at the 5 passed checks you will se that code coverage is now part of it.
If the code coverage would change by this pull request, it would also create automatically a comment stating what has changed.
I would propose that we start using it in that way and see how it goes. It also allows us to add a badge to the README, which I will do in another pull request.

frankenjoe:
So far we had covered such examples in a usage section of the documentation. You think when we move packages on GitHub we should generally have also examples in the README?

hagenw:
Yes, my idea is to have a short example  / catchy image at the README, which is also the start page of the documentation and a long usage section. We have it internally already for `audb`  this way. For `audeer` it is indeed a little bit strange, because the usage section is very short. Do you think we should not do it here, but move the new example also to the usage section?

hagenw:
Or maybe we switch the two examples and move the very short one from the usage section to the README and the longer  one to the usage section?

frankenjoe:
Yes, maybe we should switch. I like your new example, but I think it's a bit too complex as a starter.

hagenw:
I switched the two examples now.

frankenjoe:
I guess the reason you are not using `jupyter-sphinx` to run the usage example is that you don't want to introduce this dependency for only one example, right?

hagenw:
Yes, and the other reason was that it cannot display the progress bar.

PR #11, opened 2020-10-22
TST: add Windows tests

hagenw:
Add test for Windows and update the package to make it Windows compatible.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,9 +9,10 @@ on:
 jobs:
   build:
 
-    runs-on: [ ubuntu-latest, windows-latest ]
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ ubuntu-latest, windows-latest ]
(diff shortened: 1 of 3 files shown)


frankenjoe:
The coverage report only runs on `io.py` is that on purpose?

Or is it because `utils.py` is not completely covered. Because that we can simply solve if we exclude this https://github.com/audeering/audeer/blob/windows-tests/audeer/core/utils.py#L254 code block. It's anyway part of a deprecated method. Then we'd reach a `100%` code coverage.

hagenw:
The coverage report in this pull request only runs on `io.py`, because the changes are only to this part.

In general you are right we are only at 97% overall coverage as I resisted from adding a test for the deprecated method. But yes, maybe we just mark that to be excluded from coverage. I will do it in a new pull request.

frankenjoe:
Ok, should I merge with `Rebase and merge` as currently selected?

hagenw:
Strange, I selected "Squash", don't know why this reverted?

Please go with squash.

PR #14, opened 2020-10-23
TST: don't cover deprecated run_worker_threads

hagenw:
Removes deprecated `run_worker_threads` from coverage as we will not add an extra test for it.

### DIFF
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -204,7 +204,7 @@ def run_worker_threads(
         num_workers: int = None,
         progress_bar: bool = False,
         task_description: str = None
-) -> Sequence[Any]:
+) -> Sequence[Any]:  # pragma: no cover
     r"""Run parallel tasks using worker threads.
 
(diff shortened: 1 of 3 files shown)


audeer/core/utils.py
```diff
@@ -204,7 +204,7 @@ def run_worker_threads(
         num_workers: int = None,
         progress_bar: bool = False,
         task_description: str = None
-) -> Sequence[Any]:
+) -> Sequence[Any]:  # pragma: nocover
```
frankenjoe, line 207: I'm always using `pragma: no cover`, but from the test coverage we see that both works :)

frankenjoe:
I'd recommend to increase the required coverage from `90%` to `100%` now.

audeer/core/utils.py
hagenw: I did so before.
Changed it here and at another place.

hagenw:
OK, it now requires a coverage of 100%

PR #21, opened 2020-11-17
Return members of archives in extract_archive(s)

### LINKED ISSUE #20: Return extracted members
Would be nice if `audeer.extract_archive()` returns the extracted members.

hagenw:
Closes #20 

This returns a list of members for `audeer.extract_archive()` and `audeer.extract_archives()`.
Now you can do:
```python
members = audeer.extract_archive()
```

What I don't like is that the returned objects depend on the underlying archive type.
E.g. when using a ZIP file we have 
```python
members[0].filename
```
whereas when it was a TAR.GZ archive we must use
```python
members[0].name
```

Which I find unfortunate.

Should we create our own archive object to abstract from the underlying objects, or just return filenames instead.

What was your use case for `members`?

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -175,7 +175,7 @@ def extract_archives(
         *,
         keep_archive: bool = True,
         verbose: bool = False,
-) -> typing.List:
+) -> typing.List[str]:
     r"""Extract ZIP or TAR.GZ archives.
 
     Args:
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -103,7 +103,7 @@ def extract_archive(
         verbose: if ``True`` a progress bar is shown
 
     Returns:
-        members of archive
+        member filenames of archive
 
     Raises:
         RuntimeError: if the provided archive is not a ZIP or TAR.GZ file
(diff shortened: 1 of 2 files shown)


frankenjoe:
I need a list with filenames, so depending on the archive type you could change to:

```
return [member.filename for member in members]
```

and 

```
return [member.name for member in members]
```

hagenw:
I agree and think filenames are sufficient. I changed the code accordingly.

audeer/core/io.py
```diff
@@ -92,7 +92,7 @@ def extract_archive(
         *,
         keep_archive: bool = True,
         verbose: bool = False,
-):
+) -> typing.List:
```
frankenjoe, line 95:
```suggestion
) -> typing.List[str]:
```

audeer/core/io.py
```diff
@@ -161,14 +166,16 @@ def extract_archive(
...
+
 
 def extract_archives(
         archives: typing.Sequence[str],
         destination: str,
         *,
         keep_archive: bool = True,
         verbose: bool = False,
-):
+) -> typing.List:
```
frankenjoe, line 178:
```suggestion
) -> typing.List[str]:
```

PR #22, opened 2020-12-02
Add audeer.deprecated_keyword_argument decorator

hagenw:
This adds a decorator to mark keyword arguments as deprecated.

Example:
```python
import audeer


@audeer.deprecated_keyword_argument(
    deprecated_argument='foo',
    alternative_argument='bar',
    removal_version='1.0.0',
)   
def function_with_deprecated_keyword_argument(*, bar):
    return bar 
```

### DIFF
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -103,9 +103,9 @@ def deprecated_keyword_argument(
                     f"'{deprecated_argument}' argument is deprecated "
                     f"and will be removed with version {removal_version}."
                 )
+                argument_content = kwargs.pop(deprecated_argument)
                 if alternative_argument is not None:
                     message += f" Use '{alternative_argument}' instead."
-                    argument_content = kwargs.pop(deprecated_argument)
                     kwargs[alternative_argument] = argument_content
                 warnings.warn(
                     message,
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -76,7 +76,7 @@ def deprecated_keyword_argument(
     You have to specify the version,
     for which the deprecated argument will be removed.
     The content assigned to ``deprecated_argument``
-    is passed on to the ``alternative_argument``.
+    is passed on to the ``new_argument``.
 
     Args:
         deprecated_argument: keyword argument to be marked as deprecated
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -61,6 +61,62 @@ def deprecated(
     return _deprecated
 
 
+def deprecated_keyword_argument(
+        *,
+        deprecated_argument: str,
+        removal_version: str,
+        alternative_argument: str = None,
+) -> Callable:
+    r"""Mark keyword argument as deprecated.
+
+    Provide a `decorator <https://www.python.org/dev/peps/pep-0318/>`_
+    to mark keyword arguments as deprecated.
+
+    You have to specify the version,
+    for which the deprecated argument will be removed.
+    The content assigned to the deprecated keyword argument
+    is passed on to the ``new_keyword_argument``.
+
+    Args:
+        deprecated_argument: keyword argument to be marked as deprecated
+        removal_version: version the code will be removed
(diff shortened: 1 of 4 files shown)


hagenw:
What we should maybe also do after this pull request is change all our `DeprecationWarnings` to `UserWarnings`.

audeer/core/utils.py
```diff
@@ -61,6 +61,62 @@ def new_func(*args, **kwargs):
...
+        @functools.wraps(func)
+        def new_func(*args, **kwargs):
+            if deprecated_argument in kwargs:
+                message = (
+                    f"'{deprecated_argument}' argument is deprecated "
+                    f"and will be removed with version {removal_version}."
+                )
+                if alternative_argument is not None:
+                    message += f" Use '{alternative_argument}' instead."
+                    argument_content = kwargs.pop(deprecated_argument)
```
frankenjoe, line 108: Is it a good idea to only `pop` if an alternative argument is given? If this is not the case, it will probably raise an error.

frankenjoe: Another problem may arise if there is no direct mapping between the old and the new argument. Maybe we should also give the option to pass a function that would do the mapping.

hagenw: > Is it a good idea to only `pop` if an alternative argument is given? If none is provided, it will probably raise an error later.
This way we can support removal of keyword arguments. If `alternative_argument=None` this is exactly this case.

frankenjoe: But what if I want I only want to remove the keyword but not assign to an alternate argument?

hagenw: This is exactly what I'm talking about.

frankenjoe: Sorry, I'm lost. But in that case the keyword is not popped.

hagenw:
```python
@audeer.deprecated_keyword_argument(
    deprecated_argument='foo',
    alternative_argument='bar',
    removal_version='1.0.0',
) 
```
will provide an alternative keyword argument `bar` instead of `foo`.
```python
@audeer.deprecated_keyword_argument(
    deprecated_argument='foo',
    removal_version='1.0.0',
) 
```
will just remove the kyword argument `foo`

frankenjoe: I.e. the keyword is still passed on to the function, which will probably raise an error.

hagenw: Ah, ok, now I get your point. We still need to `pop` ;)

frankenjoe: Yes, that's what I meant when I asked if it is a good idea to only pop if an alternate argument is given :)

hagenw: I moved the `pop` line now before the `if`-statement.

frankenjoe: As far as I understand this decorater works  not like `deprecated` where you still have the function, but here we assume that the deprecated argument is actually removed from the parameters.

frankenjoe: > I moved the pop line now before the if-statement.

Thanks. What about an optional function that maps the old to the new argument?

hagenw: I renamed `alternative_argument` to `new_argument` and added a mapping argument, to map between the deprecated and the new argument.

audeer/core/utils.py
```diff
@@ -61,6 +61,69 @@ def new_func(*args, **kwargs):
...
+) -> Callable:
+    r"""Mark keyword argument as deprecated.
+
+    Provide a `decorator <https://www.python.org/dev/peps/pep-0318/>`_
+    to mark keyword arguments as deprecated.
+
+    You have to specify the version,
+    for which the deprecated argument will be removed.
+    The content assigned to ``deprecated_argument``
+    is passed on to the ``alternative_argument``.
```
frankenjoe, line 79:
```suggestion
    is passed on to the ``new_argument``.
```

frankenjoe:
Ok, I think we can merge this one. Please prepare another MR to change to `UserWarning' .

PR #24, opened 2020-12-03
Add function is_uid

frankenjoe:
Adds a function `is_uid` that takes as input a string and returns `True` if the string represents a unique identifier.


hagenw (approved):
Great, looks good. Your solution of first checking for a string seems to be also better than just catching all possible errors.

PR #27, opened 2021-01-27
Add audeer.version_from_git()

hagenw:
This adds a utility function that extracts a version number by the current git ref.
This will come handy for `audb.publish()` and can be used inside `docs/conf.py` to infer the current version for documentations.

It is not using a Python module to interact with git, but calls git directly which is easier and will not need another package.
If the command fails as `git` is not available or we are not inside a git repo, then `'<unknown>'` is returned as version.

Unfortunately, it is not that easy to write a test for this as it depends on the current stage of a git repo. So I repeated the code that should be executed in the test script.

image

### DIFF
--- a/audeer/__init__.py
+++ b/audeer/__init__.py
@@ -24,6 +24,7 @@ from audeer.core.utils import (
     run_worker_threads,
     to_list,
     uid,
+    version_from_git,
 )
 
 
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -449,13 +449,20 @@ def uid(
     return uid
 
 
-def version_from_git() -> str:
+def version_from_git(
+        *,
+        v=True,
+) -> str:
     r"""Get a version number from current git ref.
 
     The version is inferred executing
     ``git describe --tags --always``.
     If the command fails,
     ``'<unknown>'`` is returned.
+
+    Args:
+        v: if ``True`` version starts always with ``v``,
+            otherwise it never starts with ``v``
 
     Returns:
         version number
@@ -471,4 +478,8 @@ def version_from_git() -> str:
         version = version.decode().strip()
     except Exception:  # pragma: nocover
         version = '<unknown>'
+    if version.startswith('v') and not v:  # pragma: nocover (only local)
+        version = version[1:]
+    elif not version.startswith('v') and v:  # pragma: nocover (only github)
+        version = f'v{version}'
     return version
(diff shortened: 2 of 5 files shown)


hagenw:
For `audb.publish` I would remove the `v` at the beginning of the version strings. Should we maybe add it as an option to `audeer.version_from_git()` already?

frankenjoe:
You mean adding an argument like `as_plain_number=True`?

hagenw:
> You mean adding an argument like `as_plain_number=True`?

Yes, something like this. `as_plain_number` might be slightly misleading, what would you expect for `v1.9.0-1-g211eeeb` then?
So maybe we name it `trim_v` or `trim_starting_v`. Or we add an option `v=True`. This would then enforce `v` at the beginning and `v=False` would always remove it.

frankenjoe:
>  Or we add an option v=True. This would then enforce v at the beginning and v=False would always remove it.

I like that idea.

hagenw:
I added the `v=True` argument.

frankenjoe:
Great, works as expected.

One particular use-case is to check if the `version` corresponds to a `tag`. Is there a way to check that as well?

hagenw:
> One particular use-case is to check if the `version` corresponds to a `tag`. Is there a way to check that as well?

I'm not completely sure what the goal is here, check that the given version number is a tag?
This could be done by running:
```bash
$ git tag -l "$version"
```
but of course the problem is that we most probably have to check for both versios, one starting with `v` and the other without `v`.

Do you expect another function to check this, or integrated in the `version_from_git()` function?

frankenjoe:
> Do you expect another function to check this, or integrated in the version_from_git() function?

Either solution would be fine.

In the end it would be good to have something similar to what we have in some of our training scripts that only allow to publish a model from a tag:

```python
def current_tag(repo):
    # Get list of tags
    # https://stackoverflow.com/a/34935776
    tagmap = {}
    for tag in repo.tags:
        tagmap.setdefault(repo.commit(tag), []).append(tag)
    current_commit = repo.commit()
    if current_commit in tagmap:
        current_tags = tagmap[current_commit]
        return current_tags[0].name
    else:
        raise RuntimeError('You can only publish from a tagged commit.')
```

hagenw:
OK, but this is already one step ahead. I would also not only aallow publishing from a tag, it might be that you would like to test a version together with somebody else. But I agree that we should add a check and require maybe an extra confirmation step.

So what we would need then is `version_from_git` to get the current version, and another function `git_tags` that return a list of all available tags, for which we could also add the `v=True` option. But that I would do in another pull request.

frankenjoe:
Agree, let's move this into a new MR. Will merge now.

PR #28, opened 2021-01-29
Add audeer.git_tags()

hagenw:
As discussed in https://github.com/audeering/audeer/pull/27#issuecomment-769681087, this adds a function `git_tags()` which returns a list of all available tags.

image


frankenjoe:
Nice!

PR #31, opened 2021-02-08
Add audeer.sort_versions()

hagenw:
Add `audeer.sort_versions()` to sort semantic version numbers as they are returned e.g. by `audeer.git_repo_version()`.
It raises an error if the provided versions do not comply with `audeer.is_semantic_version()`

image


audeer/core/utils.py
```diff
@@ -1,3 +1,4 @@
+from distutils.version import LooseVersion
```
frankenjoe, line 1: If we want to keep imports sorted, this should go after `import copy`

audeer/core/utils.py
```diff
@@ -466,6 +467,39 @@ def _worker():
...
+    """Sort version numbers.
+
+    If the version starts with ``'v'``
+    it is ignored during sorting.
+
+    Args:
+        versions: sequence with semantic version numbers
+
+    Returns:
+        sorted list of version with highest as last entry
```
frankenjoe, line 482:
```suggestion
        sorted list of versions with highest as last entry
```

audeer/core/utils.py
hagenw: Good p[oint, changed it.

tests/test_utils.py
```diff
@@ -261,6 +261,56 @@ def test_run_worker_threads(func, params, expected_output):
         assert output == expected_output
 
 
+@pytest.mark.parametrize(
+    'versions, expected_versions',
+    [
+        (
+            ['1.0.0', '3.0.0', '1.10.1', '1.2.0', '1.0.1'],
```
frankenjoe, line 268: Please add a test for the empty case? And maybe one with some malformed versions, what happens in that case?

hagenw: For malformed version you mean something like `'1.2'`?

hagenw: Or `'a.b.c'`?

This is a good point, at the moment it will fail if you combine `1.2.0` with `a.b.c` as it then tries to compare integers with strings.

frankenjoe: Maybe we should add a `is_version()` function?

hagenw: I'm not completely convinced to add a `is_version()` and to restrict the allowed versions.

As a first step, I added now more tests.

`sort_versions()` will work if you use all string based versions, e.g. 'a.b.c', 'a.c.a'. It will only fail if you mix 'a.b.c' and '1.1.0'.

frankenjoe: Ok, then we should maybe add some more words to the function that explains what it can handle and when it will fail?

hagenw: Mh, I just realized the sorting will also fail for cases that Ithought should work: https://github.com/postmarketOS/pmbootstrap/issues/342
And the solution to make it work seems to complicated for me to include it here as well: https://github.com/postmarketOS/pmbootstrap/commit/19a661e98c2a7aad03673bba0ca725299c8491ad#diff-8bdac463a2857b899b00614928ac1740b70fa199c0ef02948e6ecc04c76242ac

I think, I will add a test instead that the version has to contain all three verions `X.Y.Z`, and those must be integers. Hopefully, then we are safe.

frankenjoe: Maybe in that case a function `is_version()` makes sense again :)

hagenw: OK, I created https://github.com/audeering/audeer/pull/33 as a first step.

hagenw: I'm using now `audeer.is_semantic_version()`.

Unfortunately, using `LooseVersion` is still failing for valid version numbers:

```
self = LooseVersion ('22.7.3-r1'), other = LooseVersion ('22.7.3.3-r1')
                                                   
    def _cmp (self, other):
        if isinstance(other, str): 
            other = LooseVersion(other)                   
                                                                
        if self.version == other.version:                                       
            return 0           
>       if self.version < other.version:    
E       TypeError: '<' not supported between instances of 'str' and 'int'
```

So, I'm afraid we have to add manual code as well here.

frankenjoe: :-( but then at least it's in our hands and we know exactly what is supported and what not.

hagenw: I had a look how they hanlde it at the [semver]() package:

```python
>>> semver.compare('22.7.3-r1', '22.7.3.3-r1')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-16164371ba23> in <module>
----> 1 semver.compare('22.7.3-r1', '22.7.3.3-r1')

~/.envs/test/lib/python3.6/site-packages/semver.py in wrapper(*args, **kwargs)
    157         # better remove the interpreter stack:
    158         del frame
--> 159         return func(*args, **kwargs)
    160 
    161     return wrapper

~/.envs/test/lib/python3.6/site-packages/semver.py in compare(ver1, ver2)
    855     """
    856     v1 = VersionInfo.parse(ver1)
--> 857     return v1.compare(ver2)
    858 
    859 

~/.envs/test/lib/python3.6/site-packages/semver.py in compare(self, other)
    491         cls = type(self)
    492         if isinstance(other, string_types):
--> 493             other = cls.parse(other)
    494         elif isinstance(other, dict):
    495             other = cls(**other)

~/.envs/test/lib/python3.6/site-packages/semver.py in parse(cls, version)
    724         match = cls._REGEX.match(ensure_str(version))
    725         if match is None:
--> 726             raise ValueError("%s is not valid SemVer string" % version)
    727 
    728         version_parts = match.groupdict()

ValueError: 22.7.3.3-r1 is not valid SemVer string
```

So, it seems that using a fourth `.` before the `-` is not allowed.
If we forbid that too, it will also work with `LooseVersion`:

```python
>>> audeer.sort_versions(['22.7.3-r1', '22.7.3-r1.3'])
['22.7.3-r1', '22.7.3-r1.3']
```

frankenjoe: Ok, I guess `is_semantic_version()` should fail in that case as well, right?

hagenw: Yes, I changed it accordingly in https://github.com/audeering/audeer/pull/34

hagenw: I added the fixed version of `is_semantic_version()` here, now all tests are passing.

audeer/core/utils.py
```diff
@@ -573,6 +574,52 @@ def _worker():
...
+        ...     'v2.0.0-1-gdf29c4a',
+        ... ]
+        >>> sort_versions(vers)
+        ['v1.0.0', '2.0.0', 'v2.0.0-1-gdf29c4a', '2.0.1']
+
+    """
+    for version in versions:
+        if not is_semantic_version(version):
+            raise ValueError(
+                "All version numbers have to semantic versions, "
```
frankenjoe, line 609:
```suggestion
                "All version numbers have to be semantic versions, "
```

audeer/core/utils.py
```diff
@@ -573,6 +574,52 @@ def _worker():
...
+                "following 'X.Y.Z', "
+                "where X, Y, Z are integers. "
+                f"Your vesion is instead: '{version}'."
```
frankenjoe, line 612:
```suggestion
                f"Your version is: '{version}'."
```

hagenw: I updated this manually as I also test for the error message and had to update the text there as well.

frankenjoe: Thanks!

audeer/core/utils.py
```diff
@@ -573,6 +574,52 @@ def _worker():
     return results
 
 
+def sort_versions(
+        versions: List[str],
+) -> List:
+    """Sort version numbers.
+
+    If the version starts with ``'v'``
```
frankenjoe, line 582: This sounds like versions starting with `v` are not taken into account.

hagenw: Agree, would the following work:

```rst
If the version starts with ``'v'``,
the ``'v'`` is ignored during sorting.
```

frankenjoe: Yes! But maybe rephrase to `If a version...`

hagenw: Updated now to

image

hagenw: Should I remove the `'` around the `v`?

frankenjoe: I'd say it's fine like this.

frankenjoe: Ok, actually in other places we don't put it in `'`

frankenjoe: So yes, maybe remove it then.

hagenw: Done

PR #32, opened 2021-02-08
Deprecated default value

frankenjoe:
Adds decorator `utils.deprecated_default_value()`, which raises a warning that the default value of a keyword argument will change in a future version. The warning is not raised if the value is set explicitly.


frankenjoe:
Sorry, accidentally created it from branch #31 :-(

frankenjoe:
Ok, fixed.

PR #33, opened 2021-02-08
Add audeer.is_semantic_version()

hagenw:
This adds `audeer.is_semantic_version()` to check that a given version number contains at least `X.Y.Z`, and that all those numbers are integers.

image

### DIFF
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -1,7 +1,6 @@
 import concurrent.futures
 from collections.abc import Iterable
 import copy
-from distutils.version import LooseVersion
 import functools
 import hashlib
 import queue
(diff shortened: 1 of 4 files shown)


audeer/core/utils.py
```diff
@@ -257,6 +257,52 @@ def git_repo_version(
     return version
 
 
+def is_semantic_version(version: str) -> bool:
+    r"""Check if given string represents a `semantic version`_.
```
frankenjoe, line 261: Do we have to mention that the version string can start with `v`?

hagenw: I removed that sentence from this docstring. It's shown in the eample, but it's also fine with me to introduce it back.

frankenjoe: Ok. Then let's leave it as it is.

PR #34, opened 2021-02-08
Be more strict on semantic versions

hagenw:
This now allows only for `X.Y.Z-bla` or `X.Y.Z+bla`, where `bla` can be anything and might containa `.` as well.
Added also more documentation to the doc string.

image

### DIFF
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -309,6 +309,13 @@ def git_repo_version(
 def is_semantic_version(version: str) -> bool:
     r"""Check if given string represents a `semantic version`_.
 
+    Your version is allowed to start with a ``'v'``.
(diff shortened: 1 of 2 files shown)


audeer/core/utils.py
```diff
@@ -309,6 +309,13 @@ def git_repo_version(
 def is_semantic_version(version: str) -> bool:
     r"""Check if given string represents a `semantic version`_.
 
+    Your version is allowed to start with a ``v``.
+    Otherwise it has to comply to ``X.Y.Z``,
```
frankenjoe, line 313: Here "otherwise" sounds like this only applies if it does not start with `v`. Maybe better "Apart from that"?

hagenw: I incorporated 'v' now in the second sentence:

image

PR #35, opened 2021-02-09
Correctly sort imports

frankenjoe:
Just some clean up to make sure `import xyz` are listed in sorted order and to fix some other small issues.


hagenw:
Thanks.

PR #36, opened 2021-02-19
Add audeer.create_archive()

### LINKED ISSUE #19: Function to create archive
Since we have `extract_archive()`, maybe we should also add a `create_archive()` function.

hagenw:
Closes #19 by adding a function to create ZIP and TAR.GZ archives.

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -11,6 +11,7 @@ from audeer.core.tqdm import (
     format_display_message,
     progress_bar,
 )
+from audeer.core.utils import to_list
 
 
 # Exclude common_directory example from doctest on Windows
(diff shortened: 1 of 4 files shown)


audeer/core/io.py
```diff
@@ -86,6 +87,46 @@ def all_names_equal(name):
...
+            The archive type is determined by the file extension
+
+    Raises:
+        RuntimeError: if archive does not end with ``zip`` or ``tar.gz``
+
+    """
+    archive = safe_path(archive)
+    mkdir(os.path.dirname(archive))
+    files = to_list(files)
+    if archive.endswith('zip'):
```
frankenjoe, line 113: Should we allow `ZIP`?

frankenjoe: e.g. by doing `archive.lower().endswith('zip')`?

audeer/core/io.py
```diff
@@ -86,6 +87,46 @@ def all_names_equal(name):
...
+        with zipfile.ZipFile(archive, 'w', zipfile.ZIP_DEFLATED) as zf:
+            for file in files:
+                full_file = os.path.join(root, file)
+                zf.write(full_file, arcname=file)
+    elif archive.endswith('tar.gz'):
```
frankenjoe, line 118: Same here, should we ignore case?

audeer/core/io.py
```diff
@@ -86,6 +87,46 @@ def all_names_equal(name):
...
+        with tarfile.open(archive, "w:gz") as tf:
+            for file in files:
+                full_file = os.path.join(root, file)
```
frankenjoe, line 121: should we check if file exists?

audeer/core/io.py
```diff
@@ -86,6 +87,46 @@ def all_names_equal(name):
...
+        RuntimeError: if archive does not end with ``zip`` or ``tar.gz``
+
+    """
+    archive = safe_path(archive)
+    mkdir(os.path.dirname(archive))
+    files = to_list(files)
+    if archive.endswith('zip'):
+        with zipfile.ZipFile(archive, 'w', zipfile.ZIP_DEFLATED) as zf:
+            for file in files:
+                full_file = os.path.join(root, file)
```
frankenjoe, line 116: Should we check if `full_file` exists?

audeer/core/io.py
hagenw: We also only allow for `zip` in `extract_archive`, so I would stick with lower case here.

audeer/core/io.py
hagenw: I think it should be fine if we leave this to `tarfile`

audeer/core/io.py
hagenw: I think it's fine if we leave this to `zipfile`

PR #38, opened 2021-04-13
Add basename argument to audeer.list_*_names()

### LINKED ISSUE #37: Option to return relative paths
It would be nice to have an argument in `list_dir_names()` and `list_file_names()` that returns relative paths.

hagenw:
Closes #37 

Just realized I commited changes to both functions and not only to `audeer.list_dir_names()`.
So I changed the title as it should be fine to add both in a single commit.

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -276,37 +276,43 @@ def file_extension(
 
 def list_dir_names(
         path: typing.Union[str, bytes],
+        basename: bool = False,
 ) -> typing.List:
     """List of folder names located inside provided path.
 
     Args:
         path: path to directory
+        basename: if ``True`` returns basename of directories
 
     Returns:
         list of paths to directories
 
     Example:
         >>> path = mkdir('path1/path2')
-        >>> dirs = list_dir_names('path1')
-        >>> os.path.basename(dirs[0])
-        'path2'
+        >>> list_dir_names('path1', basename=True)
+        ['path2']
 
     """
     path = safe_path(path)
     paths = [os.path.join(path, p) for p in os.listdir(path)]
-    return sorted([p for p in paths if os.path.isdir(p)])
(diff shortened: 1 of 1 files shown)


hagenw:
Should we call the argument `basenames` instead `basename` as we are also talking about `names` in the function names?

audeer/core/io.py
```diff
@@ -276,45 +276,51 @@ def file_extension(
 
 def list_dir_names(
         path: typing.Union[str, bytes],
+        basename: bool = False,
```
frankenjoe, line 279: I suggest to make `basename` a key-word only argument like in `list_file_names`

frankenjoe:
> Should we call the argument `basenames` instead `basename` as we are also talking about `names` in the function names?

yes, I think that makes sense

audeer/core/io.py
hagenw: Good catch, that was of course my plan.

I changed it accordingly.

hagenw:
> > Should we call the argument `basenames` instead `basename` as we are also talking about `names` in the function names?
> 
> yes, I think that makes sense

OK, I renamed the argument to `basenames`

audeer/core/io.py
```diff
@@ -276,45 +276,52 @@ def file_extension(
...
-        >>> os.path.basename(dirs[0])
-        'path2'
+        >>> list_dir_names('path1', basenames=True)
+        ['path2']
 
     """
     path = safe_path(path)
     paths = [os.path.join(path, p) for p in os.listdir(path)]
-    return sorted([p for p in paths if os.path.isdir(p)])
+    paths = sorted([p for p in paths if os.path.isdir(p)])
```
frankenjoe, line 299: It doesn't make a difference, but I think I would sort after the if clause.

hagenw: Yes, that's cleaner. I moved to the very end in both functions.

PR #44, opened 2021-06-01
CI: add test for macOS

### LINKED ISSUE #46: audeer.common_directory() is not working for macOS
In https://github.com/audeering/audeer/pull/44 I added tests for macOS, but they fail for `audeer.common_directoy()`:
https://github.com/audeering/audeer/pull/44/checks?check_run_id=2718638791

hagenw:
Closes #46.

Add a test for macOS and restrict Windows and MacOS tests to Python 3.8.

I also disabled the docstring test for `audeer.common_directory()` under MacOS as it puts a `/System/Volumes/Data` in front of the path in the Github runner.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,12 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, windows-latest, macOS-latest ]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8]
+        include:
+          - os: ubuntu-latest
+            python-version: 3.7
+          - os: ubuntu-latest
+            python-version: 3.9
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-latest, windows-latest ]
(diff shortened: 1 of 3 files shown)


hagenw:
There are three alternatives for the docstring example:

1. Skipping the Windows and MacOS test and stay with

image

2. Skip only the Windows test and use

image

3. Test on all platforms and write

image

frankenjoe:
I guess there is a 4. option:

```
>>> common_directory(paths).replace(os.path.sep, '/')
```

hagenw:
> I guess there is a 4. option:
> 
> ```
> >>> common_directory(paths).replace(os.path.sep, '/')
> ```

This does not work under MacOS as in the Github runner a `/System/Volumes/Data` is added at the front.

hagenw:
> Ah ok, and 3. is working because of the `...`?

Yes, exactly.
But maybe to not confuse the users, we should simply go with 1. as the code is tested in the tests anyway?

frankenjoe:
> But maybe to not confuse the users, we should simply go with 1. as the code is tested in the tests anyway?

Yes, and the official documentation we build on Ubuntu, so the user will most likely not notice we skip the tests on Windows and MacOS.

hagenw:
> > But maybe to not confuse the users, we should simply go with 1. as the code is tested in the tests anyway?
> 
> Yes, and the official documentation we build on Ubuntu, so the user will most likely not notice we skip the tests on Windows and MacOS.

I restricted the docstring example now to Linux as before.

PR #48, opened 2021-11-19
Add audeer.rmdir()

hagenw:
As I don't like to import `shutil` everytime I need to delete a directory and also first check if the directory exists, I decided to add `audeer.rmdir()`.

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -474,7 +474,7 @@ def rmdir(
         NotADirectoryError: if path is not a directory
 
     Example:
-        >>> p = mkdir('path1/path2/path3')
+        >>> mkdir('path1/path2/path3')  # doctest: +SKIP
         >>> rmdir('path1/path2')
         >>> list_dir_names('path1')
         []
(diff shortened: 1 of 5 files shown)


audeer/core/io.py
```diff
@@ -458,6 +458,33 @@ def replace_file_extension(
...
+    if the directory exists.
+
+    Args:
+        path: absolute or relative path of directory to remove
+
+    Raises:
+        NotADirectoryError: if path is not a directory
+
+    Example:
+        >>> p = mkdir('path1/path2/path3')
```
frankenjoe, line 477: why assigning to `p`?

hagenw: I assign to a variable as otherwise the test will fail.
So maybe better
```suggestion
        >>> _ = mkdir('path1/path2/path3')
```
?

hagenw: At the beginning I thought about reusing the returned path, but `os.path.dirname(p)` is much harder to understand than just `path1/path2`

frankenjoe: Alternatively you could use `# doctest: +SKIP`, but I'm also fine with your solution.

hagenw: Yeah, good suggestion as we test this line anyway by the next lines and it looks cleaner without the return variable.
I changed to use `# doctest: +SKIP`

PR #49, opened 2021-11-24
Add verbose flag to create_archive()

frankenjoe:
We already had a verbose argument to show a progress bar in `extract_archive()`. This adds one to `create_archive()`. It also simplifies the code where we create the progress bar in `extract_archive()`.

image


hagenw:
Thanks.
I tested both progress bars locally, and everything worked as expected.

PR #52, opened 2022-03-01
Add utils.install_package()

frankenjoe:
image


frankenjoe:
And of course the new tests are failing on Windows.. this is so sad :-(

frankenjoe:
Interestingly it fails only when trying to uninstall the package:

```
 E           subprocess.CalledProcessError: Command '['C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\python.exe', '-m', 'pip', 'uninstall', '--yes', 'PyYaml']' returned non-zero exit status 2.
```

frankenjoe:
Ẁeird enough it works in https://github.com/audeering/audobject/pull/35, but there we use Python 3.8 instead of 3.7

frankenjoe:
Mhh, maybe it has to do with the `PyYaml` package. Something seems broken in the path, it looks in a directory `~aml`.

```
PermissionError: [WinError 5] Access is denied: 'C:\\hostedtoolcache\\windows\\Python\\3.9.10\\x64\\Lib\\site-packages\\~aml\\_yaml.cp39-win_amd64.pyd'
```

frankenjoe:
Ok, Windows tests are passing when switching from `PyYaml`  to `audbackend`. 

But now there is a new issue, when trying to verify the version we get, see e.g. https://github.com/audeering/audeer/runs/5373327135?check_suite_focus=true#step:5:53

```

    # import module and verify version
    m = importlib.import_module(MODULE)
    if version is not None:
        m.__version__ == version
```
```
 E           AttributeError: module 'audbackend' has no attribute '__version__'
```

Locally it works without problems.

@hagenw any idea why?

audeer/core/utils.py
```diff
@@ -306,6 +308,44 @@ def git_repo_version(
     return version
 
 
+def install_package(
+        name: str,
+        *,
+        version: str = None,
+        silent: bool = False,
+):
+    r"""Install pip package.
```
hagenw, line 317:
```suggestion
    r"""Install a Python package with pip.
```

hagenw:
image

That's interesting. Locally I get

```python
>>> import audbackend
>>> audbackend.__version__
'0.3.12'
```

Does it mean it is not installed properly in the tests here?

hagenw:
Oh, sorry, now I see you ask the same question ;)

hagenw:
I guess we need to debug. So maybe add a `print(m)` statement before the `m.__version__` line as a first step.

frankenjoe:
There is another problem. If a version of the package already exists, but the user requests a different version, it will uninstall the current version and install the new version. Now consider the following example (we start from an environment where `audbackend` is not installed):

```python
audeer.install_package(
    'audbackend',
    version='0.3.12',
    silent=False,
)

m = importlib.import_module('audbackend')
print(m.__version__)
m.Repository('name', 'host', 'backend')
```
```
Looking in indexes: https://artifactory.audeering.com/artifactory/api/pypi/pypi/simple
Collecting audbackend==0.3.12
  Using cached https://artifactory.audeering.com/artifactory/api/pypi/pypi/packages/packages/fb/7f/0a3dcbbcc17f0d400089dbc94a4ffe2cd78266c7f9a320560a2439ed292c/audbackend-0.3.12-py3-none-any.whl (10 kB)
Requirement already satisfied: audeer>=1.17.1 in /media/jwagner/Data/Git/pyaudeer (from audbackend==0.3.12) (1.17.3.dev9+gc88f79c)
Requirement already satisfied: audfactory>=1.0.3 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from audbackend==0.3.12) (1.0.8)
Requirement already satisfied: dohq-artifactory==0.7.742 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from audbackend==0.3.12) (0.7.742)
Requirement already satisfied: requests in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from dohq-artifactory==0.7.742->audbackend==0.3.12) (2.24.0)
Requirement already satisfied: PyJWT in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from dohq-artifactory==0.7.742->audbackend==0.3.12) (2.3.0)
Requirement already satisfied: python-dateutil in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from dohq-artifactory==0.7.742->audbackend==0.3.12) (2.8.2)
Requirement already satisfied: tqdm in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from audeer>=1.17.1->audbackend==0.3.12) (4.50.2)
Requirement already satisfied: six>=1.5 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from python-dateutil->dohq-artifactory==0.7.742->audbackend==0.3.12) (1.15.0)
Requirement already satisfied: chardet<4,>=3.0.2 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory==0.7.742->audbackend==0.3.12) (3.0.4)
Requirement already satisfied: certifi>=2017.4.17 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory==0.7.742->audbackend==0.3.12) (2020.6.20)
Requirement already satisfied: urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory==0.7.742->audbackend==0.3.12) (1.25.11)
Requirement already satisfied: idna<3,>=2.5 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory==0.7.742->audbackend==0.3.12) (2.10)
Installing collected packages: audbackend
Successfully installed audbackend-0.3.12
0.3.12
```

Now, we install '0.3.6':

```python
audeer.install_package(
    'audbackend',
    version='0.3.6',
    silent=False,
)

m = importlib.import_module('audbackend')
print(m.__version__)
m.Repository('name', 'host', 'backend')
```
```
Looking in indexes: https://artifactory.audeering.com/artifactory/api/pypi/pypi/simple
Collecting audbackend==0.3.6
  Using cached https://artifactory.audeering.com/artifactory/api/pypi/pypi/packages/packages/f0/41/389782c9e1d19bb8c31ebf5d6b83ea85c2aca82820937911ea94f90c7010/audbackend-0.3.6-py3-none-any.whl (9.8 kB)
Requirement already satisfied: audfactory>=1.0.3 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from audbackend==0.3.6) (1.0.8)
Requirement already satisfied: audeer>=1.12.0 in /media/jwagner/Data/Git/pyaudeer (from audbackend==0.3.6) (1.17.3.dev9+gc88f79c)
Requirement already satisfied: tqdm in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from audeer>=1.12.0->audbackend==0.3.6) (4.50.2)
Requirement already satisfied: dohq-artifactory>=0.7.377 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from audfactory>=1.0.3->audbackend==0.3.6) (0.7.742)
Requirement already satisfied: PyJWT in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (2.3.0)
Requirement already satisfied: requests in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (2.24.0)
Requirement already satisfied: python-dateutil in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (2.8.2)
Requirement already satisfied: six>=1.5 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from python-dateutil->dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (1.15.0)
Requirement already satisfied: chardet<4,>=3.0.2 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (3.0.4)
Requirement already satisfied: idna<3,>=2.5 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (2.10)
Requirement already satisfied: certifi>=2017.4.17 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (2020.6.20)
Requirement already satisfied: urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1 in /media/jwagner/Data/Git/pyaudeer/venv/lib/python3.6/site-packages (from requests->dohq-artifactory>=0.7.377->audfactory>=1.0.3->audbackend==0.3.6) (1.25.11)
Installing collected packages: audbackend
  Attempting uninstall: audbackend
    Found existing installation: audbackend 0.3.12
    Uninstalling audbackend-0.3.12:
      Successfully uninstalled audbackend-0.3.12
Successfully installed audbackend-0.3.6
0.3.12
```

The output on the command line tells us that '0.3.12' was uninstalled and we have now '0.3.6'. But when we print the version it's still '0.3.12' and also `Repository` was only added after '0.3.6', but no error is raised.

The changes only take place after stopping the process. Afterwards, when I run the following code I get the expected output.

```python
m = importlib.import_module('audbackend')
print(m.__version__)
m.Repository('name', 'host', 'backend')
```
```
0.3.6
...
AttributeError: module 'audbackend' has no attribute 'Repository'
```

So there seems to some caching involved.

frankenjoe:
According to this comment from https://stackoverflow.com/questions/2918898/prevent-python-from-caching-the-imported-modules there is quite some danger in reloading of a module:

image

So I think we have to be careful here here. At the moment I see two solutions: either not implement the function in `audeer` or  restrict it only install a package if no version is installed yet.

@hagenw opinion?

hagenw:
Yes, as there seems to be not a way to implement this in a meaningful way for a general purpose, I would also suggest to not add it here, but only to `audobject`.
BTW, there it can also happen that a different version of a package is already installed (e.g. `opensmile`) and then `audobject` updates it when loading a YAML file, so we can also have the same problems there.

frankenjoe:
> BTW, there it can also happen that a different version of a package is already installed (e.g. `opensmile`) and then `audobject` updates it when loading a YAML file, so we can also have the same problems there.

No, because we only install the package if the module cannot be loaded. This means that it can happen that you have a too old version of the package installed, so we probably need to extend the documentation saying a package is only installed if the module is not yet installed.

hagenw:
> > BTW, there it can also happen that a different version of a package is already installed (e.g. `opensmile`) and then `audobject` updates it when loading a YAML file, so we can also have the same problems there.
> 
> No, because we only install the package if the module cannot be loaded. Of course it means that it can happen that you have a too old version of the package installed, so we probably need to extend the documentation saying a package is only installed if the module is not yet installed.

I guess we could check for the correct version.

You could also extend `audeer.install_package()` to first check if the package is already installed, and if it has the wrong version raise an error and state that it needs to be uninstalled first.

frankenjoe:
> I guess we could check for the correct version.

I think in `audobject` we already raise a warning if an older of the package is installed (need to check though, maybe we only do so if an argument is not found). But in `audobject` I would not raise an error I think. It may still work with an older version and maybe there is a reason the user has to have an older version installed.

> You could also extend audeer.install_package() to first check if the package is already installed, and if it has the wrong version raise an error and state that it needs to be uninstalled first.

Mhh ok, that could be an option indeed. And in case a newer version is installed we do nothing?

frankenjoe:
Ah sorry, you suggest to always raise an error, even if a newer version is installed. Makes sense of course.

hagenw:
> Or maybe we add an argument whether an error should be raised if a newer version is already available.

As you would like to allow also for older packages in `audobject`, I would maybe add an option, if an error should be raised if a different version than the specified one is installed.

BTW, how do you handle versions like `>=1.0.0`. I guess this is not supported and I have to provide `1.0.0`. if yes do you raise an error for `>=1.0.0`?

frankenjoe:
> BTW, how do you handle versions like `>=1.0.0`. I guess this is not supported and I have to provide `1.0.0`. if yes do you raise an error for `>=1.0.0`?

Currently looking into this. It would actually be a nice solution if the user can use `>=1.0.0` and in that case we would only raise an error if a version below `1.0.0` is installed. Otherwise keep the current version or if it is not installed, install the latest.

frankenjoe:
I have now implemented support for `>`, `>=`, `<`, `<=`, otherwise an error is raised if the package is already installed. See description. Locally it works, but with `Python >= 3.6` it now raises:

```
E               pkg_resources.ContextualVersionConflict: (audeer 0.1.dev1+g91234cb (/home/runner/work/audeer/audeer), Requirement.parse('audeer>=1.12.0'), {'audbackend'})
```

Quite strange that `audeer` has version `0.1.dev1+g91234cb`. Anyway, to solve the error we will have to switch to a package that does not depend on `audeer` I guess.

frankenjoe:
Ok, I switched to `PyYaml` and now the tests are finally passing! What a struggle :)

audeer/core/utils.py
```diff
@@ -4,15 +4,20 @@
...
+import pkg_resources
 import queue
 import subprocess
+import sys
 import threading
 import typing
 import uuid
 import warnings
 
+import audeer
```
hagenw, line 20:
```suggestion
```

frankenjoe:
> Can you update the screenshot in the description, please.

done

audeer/core/utils.py
```diff
@@ -306,6 +311,105 @@ def git_repo_version(
...
+    try:
+        current_version = pkg_resources.get_distribution(name).version
+    except pkg_resources.DistributionNotFound:
+        current_version = None
+
+    if current_version is not None:
+        if version is None:
+            return  # any version is fine
+        if op(LooseVersion(current_version), LooseVersion(version)):
+            return  # installed version satisfies requested version
```
hagenw, line 378: Why are you not checking for `==` here as well?

tests/test_install.py
```diff
@@ -0,0 +1,103 @@
...
+            '-m',
+            'pip',
+            'uninstall',
+            '--yes',
+            PACKAGE,
+        ],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+    # remove from module cache
```
hagenw, line 27: Would this also work in general?
If yes we would have the possibility to first remove the package if a wrong version of it is installed.

tests/test_install.py
```diff
@@ -0,0 +1,103 @@
...
+        PACKAGE,
+        version='>=5.3',
+    )
+    audeer.install_package(
+        PACKAGE,
+        version='>5.2',
+    )
+    audeer.install_package(
+        PACKAGE,
+        version='<=5.3',
```
hagenw, line 72: Why this test. It is identical to the first installation, and all the commands in between should not have changed anything?

audeer/core/utils.py
frankenjoe: We do, the default value of `op` is `eq`.

audeer/core/utils.py
frankenjoe: I have no idea how this happened :) I removed it.

audeer/core/utils.py
hagenw: Ah, yes, now I see it.

audeer/core/utils.py
```diff
@@ -306,6 +310,107 @@ def git_repo_version(
...
+            "does not satisfy "
+            f"'{version_org}'."
+        )
+
+    # install package
+    version = version_org
+    if version is not None:
+        if op == operator.eq:
+            # since we do not support ==1.0
+            # we have add it here
```
hagenw, line 392:
```suggestion
            # we have to add it here
```

PR #54, opened 2022-03-01
Add audeer.move_file()

hagenw:
As proposed in https://github.com/audeering/audb/pull/175#discussion_r813996803 this adds `audeer.move_file()`.

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -439,8 +439,8 @@ def mkdir(
 
 
 def move_file(
-        src_file,
-        dst_file,
+        src_path,
+        dst_path,
 ):
     """Move a file independent of operating system.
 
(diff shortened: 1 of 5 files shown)


audeer/core/io.py
```diff
@@ -438,6 +438,35 @@ def mkdir(
...
+    we use :func:`os.replace`
+    to move the file.
+
+    Args:
+        scr_path: source file path
+        dst_path: destination file path
+
+    Example:
+        >>> path1 = mkdir('path1')
+        >>> path2 = mkdir('path1/path2')
```
frankenjoe, line 459:
```suggestion
        >>> path2 = mkdir(os.path.join('path1', 'path2'))
```

frankenjoe: Actually I wonder if we should simplify the example and just move a file within the same directory.

hagenw: The problem is I cannot use `path` here as this will break other doctests,
but I switched to a single directory named `folder` now.

hagenw: done.

PR #56, opened 2022-03-02
Add audeer.LooseVersion() and audeer.StrictVersion()

### LINKED ISSUE #55: Replace distutils.version.LooseVersion
`distutils.version` is deprecated and will be removed in the future.

There is an alternative available with the [packaging](https://github.com/pypa/packaging) package, but this would mean to introduce a dependency to `packaging` and `pyparsing`, which is not a nice solution.

So far I haven't found a better alternative though.

Which currently leaves us with the following alternatives:

1. Copy the code from `distutils.version`
2. Copy the code from `packaging.version` (which does not depend on external packages, compare https://github.com/pypa/packaging/blob/main/packaging/version.py)
3. Implement ourselves a function that can compare version numbers

At the moment we use `LooseVersion` in `audeer.sort_versions()`, `audb.load()`.

@frankenjoe any preference?

hagenw:
Closes #55

This adds `audeer.LooseVersion` and `audeer.StrictVersion` copying the implementation of `distutils.version.LooseVersion` from https://github.com/python/cpython/blob/20a1c8ee4bcb1c421b7cca1f3f5d6ad7ce30a9c9/Lib/distutils/version.py

image

image

### DIFF
--- a/audeer/__init__.py
+++ b/audeer/__init__.py
@@ -38,7 +38,8 @@ from audeer.core.utils import (
     uid,
 )
 from audeer.core.version import (
-    Version,
+    LooseVersion,
+    StrictVersion,
 )
 
 
--- a/audeer/__init__.py
+++ b/audeer/__init__.py
@@ -37,6 +37,9 @@ from audeer.core.utils import (
     to_list,
     uid,
 )
+from audeer.core.version import (
+    Version,
+)
 
 
 # Discourage from audeer import *
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -1,7 +1,6 @@
 import concurrent.futures
 from collections.abc import Iterable
 import copy
-from distutils.version import LooseVersion
 import functools
 import hashlib
 import importlib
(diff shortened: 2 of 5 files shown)


frankenjoe:
> I decided to go only for `LooseVersion` and skip `StrictVersion` as I don't think we will need it.

I think it would be useful to include both. At the moment we only have `is_semantic_version()` to check for `x.y.z`, which returns `False` for `x.y`, which e.g. some packages `pip` packages use. So I think we should also add a function `is_strict_version()` or something like this. We don't have to do it in this PR, but we should already keep `StrictVersion` I think.

hagenw:
> > I decided to go only for `LooseVersion` and skip `StrictVersion` as I don't think we will need it.
> 
> I think it would be useful to include both. At the moment we only have `is_semantic_version()` to check for `x.y.z`, which returns `False` for `x.y`, which e.g. some packages `pip` packages use. So I think we should also add a function `is_strict_version()` in addition. We don't have to do it in this PR, yet it would be useful to keep `StrictVersion` I think.

OK, I changed it and added now both `StrictVersion` and `LooseVersion`.

hagenw:
I still need to rebase the master before we can merge here.

hagenw:
> I still need to rebase the master before we can merge here.

Done.

PR #57, opened 2022-03-02
list_file_names(): add recursive argument

frankenjoe:
image


audeer/core/io.py
```diff
@@ -363,37 +363,51 @@ def list_file_names(
...
         basenames: bool = False,
-) -> typing.List:
+        recursive: bool = False,
+) -> typing.List[str]:
     """List of file names inferred from provided path.
 
     Args:
         path: path to file, directory or pattern
         filetype: optional consider only this filetype
         basenames: if ``True`` returns basenames of directories
```
hagenw, line 373:
```suggestion
        basenames: if ``True`` return relative path in respect to ``path``
```

hagenw:
After we merge this, you should then also add it to `list_dir_names()`

frankenjoe:
> After we merge this, you should then also add it to `list_dir_names()`

exactly

PR #58, opened 2022-03-02
Possibility to join paths in audeer.safe_path()

### LINKED ISSUE #18: Combine safe_path with os.path.join
In my code I have a lot of these statements:

```
path = audeer.safe_path(os.path.join('/root',  'sub', 'my.txt'))
```

would be nice if I could simply do:

```
path = audeer.safe_path('/root',  'sub', 'my.txt')
```

hagenw:
Closes #18.

This expands `audeer.safe_path()` by adding an optional `*paths` argument, which allows to provide parts of a path, which are then first joined by `os.path.join()`.

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -544,11 +544,12 @@ def safe_path(
 
     Args:
         path: path to file, directory
-        *paths: if additional arguments are provided,
-            they are first joined with :func:`os.path.join`
+        *paths: additional arguments
+            to be joined with ``path``
+            by :func:`os.path.join`
 
     Returns:
-        expanded path
+        (joined and) expanded path
 
     Example:
         >>> home = safe_path('~')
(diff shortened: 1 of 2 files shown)


hagenw:
Maybe we should have named `audeer.safe_path()` only `audeer.path()`.

But we can also still add `audeer.path()`, and let `audeer.safe_path()` copy it.
I would not directly deprecate `audeer.safe_path` for version 2.0.0 as it would generate deprecations for all our code.
I'm also not sure if I would remove `audeer.safe_path()` directly from the documentation. Maybe first leave it in and just state that it is a copy of `audeer.path()`, which should be used instead.

frankenjoe:
> This expands `audeer.safe_path()` by adding an optional `*paths` argument, which allows to provide parts of a path, which are then first joined by `os.path.join()`.

Cool, great feature!

> Maybe we should have named audeer.safe_path() only audeer.path().

Good idea. And yes, we don't have to deprecate it but just call `audeer.path()` under the hood.

PR #60, opened 2022-03-02
Update license year

hagenw:
Updates the year in the license file.


frankenjoe:
I guess this is something we have to do in all our projects. We cannot simply say: `Copyright (c) 2018-present audEERING GmbH and Contributors` right?

hagenw:
A nice idea, `present` is also used in other LICENSE files, compare https://github.com/pytorch/pytorch/blob/master/LICENSE#L15. I will change accordingly.

PR #61, opened 2022-03-02
Add audeer.path()

hagenw:
As suggested in https://github.com/audeering/audeer/pull/58#issuecomment-1056896331 this adds `audeer.path()`.
It also extends the docstring compared to what we had before in `audeer.safe_path()`.

image

I added a warning to `audeer.safe_path()` in the documentation:

image

### DIFF
--- a/audeer/__init__.py
+++ b/audeer/__init__.py
@@ -11,12 +11,12 @@ from audeer.core.io import (
     list_file_names,
     mkdir,
     move_file,
-    path,
     replace_file_extension,
     rmdir,
     safe_path,
     touch,
 )
+from audeer.core.path import path
 from audeer.core.tqdm import (
     format_display_message,
     progress_bar,
--- a/audeer/__init__.py
+++ b/audeer/__init__.py
@@ -13,10 +13,12 @@ from audeer.core.io import (
     move_file,
     replace_file_extension,
     rmdir,
-    safe_path,
     touch,
 )
-from audeer.core.path import path
+from audeer.core.path import (
+    path,
+    safe_path,
+)
 from audeer.core.tqdm import (
     format_display_message,
     progress_bar,
--- a/audeer/__init__.py
+++ b/audeer/__init__.py
@@ -11,6 +11,7 @@ from audeer.core.io import (
     list_file_names,
     mkdir,
     move_file,
+    path,
     replace_file_extension,
     rmdir,
     safe_path,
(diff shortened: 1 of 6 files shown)


hagenw:
It does not update much of the code inside `audeer/core/io.py` to avoid having conflicts between the function name `path()` and the argument `path` used at different places.

Also to make sure that `audeer.safe_path()` is also covered, I did not use `audeer.path()` in the tests.

frankenjoe:
There are still quite a few uses of `safe_path()` in other `io` functions, should we switch to `path()` here? And maybe put `_path = path` and an additional `path = _path` only around the `safe_path()` function?

hagenw:
> There are still quite a few uses of `safe_path()` in other `io` functions, should we switch to `path()` here? And maybe put `_path = path` and an additional `path = _path` only around the `safe_path()` function?

We cannot easily switch to use `path()` as we always use `path` as argument variable name which would then hide `path()`.
I moved `path()` now to `audeer/core/path.py` and import in inside `audeer/core/io.py` as `_path` which is only used inside `safe_path()`. When we at some point delete `safe_path()` from `audeer/core/io.py` we can then do
```python
from audeer.core.path import path as safe_path
```

hagenw:
Maybe I should also move `safe_path()` to `audeer/core/path.py` then we could do the 

```python
from audeer.core.path import path as safe_path
```

already inside `audeer/core/io.py` without the need to use the actual `safe_path()` function, which could then raise a deprecation warning somewhere in the future.

frankenjoe:
Great, I think that's a good solution.

frankenjoe:
And you think we really need to include `safe_path()` in the documentation?

hagenw:
> Great, I think that's a good solution.

Changed it accordingly.

hagenw:
> And you think we really need to include `safe_path()` in the documentation?

Yes, I think we should have it there for the next couple of month.
Because a user might read code, see `audeer.safe_path()` and wonder what it is doing.

frankenjoe:
> Because a user might read code, see `audeer.safe_path()` and wonder what it is doing.

Ok! But think after moving it to `path.py` you now need a `# pragma: no cover`.

hagenw:
> > Because a user might read code, see `audeer.safe_path()` and wonder what it is doing.
> 
> Ok! But think after moving it to `path.py` you now need a `# pragma: no cover`.

I added `safe_path()` to the tests.

PR #62, opened 2022-03-02
list_dir_names(): add recursive argument

frankenjoe:
image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -345,7 +345,7 @@ def list_dir_names(
 
     Args:
         path: path to directory
-        basenames: if ``True`` returns basenames of directories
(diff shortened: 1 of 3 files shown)


audeer/core/io.py
```diff
@@ -338,27 +339,41 @@ def list_dir_names(
...
         *,
         basenames: bool = False,
-) -> typing.List:
+        recursive: bool = False,
+) -> typing.List[str]:
     """List of folder names located inside provided path.
 
     Args:
         path: path to directory
         basenames: if ``True`` returns basenames of directories
```
hagenw, line 348: I would also change it here like we did for `audeer.list_file_names()`:

```suggestion
        basenames: if ``True`` return relative path in respect to ``path``
```

PR #67, opened 2022-04-08
Add argument to include hidden files

### LINKED ISSUE #66: audeer.list_file_names() ignores hidden files
As the following example shows `audeer.list_file_names()` does not return hidden files, but `audeer.list_dir_names()` does.

```python
with tempfile.TemporaryDirectory() as root:
    audeer.touch(audeer.path(root, 'file'))
    audeer.touch(audeer.path(root, '.file'))
    audeer.mkdir(audeer.path(root, 'folder'))
    audeer.mkdir(audeer.path(root, '.folder'))
    files = audeer.list_file_names(root, basenames=True)
    print(files)
    dirs = audeer.list_dir_names(root, basenames=True)
    print(dirs)
```
```
['file']
['.folder', 'folder']
```

frankenjoe:
Closes #66 

To be backward compatible defaults to `False` for `audeer.list_file_names()` and to `True` for `audeer.list_dir_names()`.

image

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -1,4 +1,4 @@
-from glob import glob
+import fnmatch
 import itertools
 import os
 import platform
@@ -340,6 +340,7 @@ def list_dir_names(
         *,
         basenames: bool = False,
         recursive: bool = False,
+        hidden: bool = True,
 ) -> typing.List[str]:
     """List of folder names located inside provided path.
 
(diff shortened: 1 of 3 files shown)


audeer/core/io.py
```diff
@@ -340,30 +340,48 @@ def list_dir_names(
...
         recursive: bool = False,
+        hidden: bool = True,
 ) -> typing.List[str]:
     """List of folder names located inside provided path.
 
     Args:
         path: path to directory
         basenames: if ``True`` return relative path in respect to ``path``
         recursive: if ``True`` includes subdirectories
+        hidden: if ``True`` includes directories starting with a dot (.)
```
hagenw, line 351:
```suggestion
        hidden: if ``True`` includes directories starting with a dot (``.``)
```

audeer/core/io.py
```diff
@@ -391,43 +410,96 @@ def list_file_names(
         filetype: optional consider only this filetype
         basenames: if ``True`` return relative path in respect to ``path``
         recursive: if ``True`` includes subdirectories
+        hidden: if ``True`` includes files starting with a dot (.)
```
hagenw, line 413:
```suggestion
        hidden: if ``True`` includes files starting with a dot (``.``)
```

hagenw:
Under Windows hidden files/folders do not seem to be handled via `.`, but with some special attribute flag, compare https://stackoverflow.com/a/14063074.
But I would not try to support it and stick with `.` as we also clearly state this in the docstring.

tests/test_io.py
```diff
@@ -192,18 +192,39 @@ def test_file_extension(path, extension):
...
+        (['a'], ['a'], True, False),
+        (
+            ['a', os.path.join('a', 'b'), os.path.join('a', 'b', 'c')],
+            ['a', os.path.join('a', 'b'), os.path.join('a', 'b', 'c')],
+            True,
+            False,
+        ),
+        # hidden
+        (['a', '.b'], ['a'], True, False),
+        (['a', '.b'], ['.b', 'a'], True, True),
```
hagenw, line 211: I guess this means hidden folders are always placed at the beginning in the sorted returned path list, correct?

frankenjoe: Yes, it's because we sort the files at the end. But maybe we should first sort the non-hidden files and then append the hidden files so that instead of:

  ```
.file.wav
file.wav
```

we return:

```
file.wav
.file.wav
```

The other question is, it makes sense to only sort within sub-directories, i.e. instead of:

```
a/file.wav
file.wav
```

we return:

```
file.wav
a/file.wav
```

PR #70, opened 2022-06-24
Add remove_from_kwargs argument to deprecated_keyword_argument()

### LINKED ISSUE #69: deprecated_keyword_argument does not always provide access to the content of the argument
If you want to deprecate a keyword argument, but you will not replace it by a different argument you will loose access to the content of the argument as it is removed from `kwargs`.

I would suggest to only remove it from `kwargs` if a new argument is defined.

hagenw:
Closes #69 

This adds the keyword argument `remove_from_kwargs` to `audeer.deprecated_keyword_argument()` with default setting of `True`, which was the behavior before. If you set it to `False` it will not remove the deprecated keyword argument from `kwargs`.

image

### DIFF
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -123,6 +123,7 @@ def deprecated_keyword_argument(
         removal_version: str,
         new_argument: str = None,
         mapping: typing.Callable = None,
+        remove_from_kwargs: bool = True,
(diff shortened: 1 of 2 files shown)


hagenw:
@frankenjoe I think we forgot about this one. It looks still ready for review.

Issue #87, opened 2023-01-26
Replace with empty file extension


frankenjoe:
When we replace with an empty extension we currently get:

```python
>>> import audeer
>>> file = 'foo.bar'
>>> audeer.replace_file_extension(file, '')
'foo.'
>>> audeer.replace_file_extension(file, '', ext='.bar')
'foo.'
>>> audeer.replace_file_extension(file, '', ext='bar')
'foo.'
```

I wonder if makes more sense if we remove the '.' in that case?

hagenw:
I guess yes, as you would expect to don't have a `.` when you remove the file extension.

hagenw:
We now remove the `'.'` which makes sense, so I will close here.

PR #90, opened 2023-01-26
Fix audeer.replace_file_extension() for empty extensions

### LINKED ISSUE #89: Replace file extension on a file without extension
Currently if a file does not have an extension the file name is replaced:

```python
>>> file = 'foo'
>>> audeer.replace_file_extension(file, 'bar')
bar
```

But maybe it makes more sense to return `foo.bar` in that case?

frankenjoe:
Closes #89 #87 

Fixes #89 and #87 ~~and raises an error if `path` does not end on `ext`~~. For consistency with `ext`, `new_extension` may now also start with a dot.

image

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -611,7 +611,7 @@ def test_move_file(tmpdir, src_file, dst_file):
     'path, new_extension, ext, expected_path',
     [
         ('', '', None, ''),
-        ('', 'txt', None, '.txt'),
+        ('', 'txt', None, ''),
         ('file', '', None, 'file'),
         ('file', 'txt', None, 'file.txt'),
         ('file.txt', 'wav', None, 'file.wav'),
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -621,7 +621,6 @@ def test_move_file(tmpdir, src_file, dst_file):
         ('file.txt', 'wav', '.txt', 'file.wav'),
         ('file.txt', '.wav', 'txt', 'file.wav'),
         ('file.txt', '.wav', '.txt', 'file.wav'),
-        ('file.txt', 'wav', 't', 'file.txwav'),
         ('file.a.b', 'wav', 'a.b', 'file.wav'),
         ('file.a.b', 'wav', '.a.b', 'file.wav'),
         pytest.param(
@@ -635,6 +634,13 @@ def test_move_file(tmpdir, src_file, dst_file):
             'file.txt',
             'wav',
             'bad',
+            None,
+            marks=pytest.mark.xfail(raises=RuntimeError),
+        ),
+        pytest.param(
+            'file.txt',
+            'wav',
+            't',
             None,
             marks=pytest.mark.xfail(raises=RuntimeError),
         ),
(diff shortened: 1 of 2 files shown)


frankenjoe:
Currently, we have the following:

```python
>>> audeer.replace_file_extension('.foo', 'bar')
.foo.bar
```

first I thought the latter should return `.bar`, but then I realized that the following returns an empty string:

```python
>>> audeer.file_extension('.foo')

```

which is in-line with:

```python
>>> os.path.splitext('.foo')
('.foo', '')
```

So I think `'.foo.bar'` is the correct result.

But with this in mind, the following output is problematic:

``` python
>>> audeer.replace_file_extension('', 'bar')
.bar
```

and I wonder if we should rather raise an error in that case?

hagenw:
> raises an error if path does not end on ext

This I would not change as it will break a lot of old scripts.
When I use `audeer.replace_file_extension()` on a bunch of files and would like to replace only files ending with `ext1`, I would use `ext='ext1'` which would have replaced only the desired files and skipped the others, whereas now it raises an error for them.

hagenw:
> But with this in mind, the following output is problematic:
> 
> ```python
> >>> audeer.replace_file_extension('', 'bar')
> .bar
> ```
> 
> and I wonder if we should rather raise an error in that case?

Another solution would be:

```python
>>> audeer.replace_file_extension('', 'bar')
''
```

hagenw:
We support the following (is part of the tests):

```python
>>> audeer.replace_file_extension('test.txt', 'wav', ext='t')
'file.txwav'
```
But the following feels a little bit counter-intuitive to me:

```python
>>> audeer.replace_file_extension("test.txt", "wav", ext=".t")
'test.txwav'
>>> audeer.replace_file_extension("test.txt", ".wav", ext="t")
'test.txwav'
```

frankenjoe:
> Another solution would be:
> 
> ```python
> >>> audeer.replace_file_extension('', 'bar')
> ''
> ```

Good idea, changed it accordingly.

frankenjoe:
> We support the following (is part of the tests):
> 
> ```python
> >>> audeer.replace_file_extension('test.txt', 'wav', ext='t')
> 'file.txwav'
> ```
> 
> But the following feels a little bit counter-intuitive to me:
> 
> ```python
> >>> audeer.replace_file_extension("test.txt", "wav", ext=".t")
> 'test.txwav'
> >>> audeer.replace_file_extension("test.txt", ".wav", ext="t")
> 'test.txwav'
> ```

You are right, I changed to raise an exception when it does not match the full extension.

frankenjoe:
> This I would not change as it will break a lot of old scripts.
> When I use `audeer.replace_file_extension()` on a bunch of files and would like to replace only files ending with `ext1`, I would use `ext='ext1'` which would have replaced only the desired files and skipped the others, whereas now it raises an error for them.

Oh, just saw your comment now. Ok, i.e. if it does match the extension we return the original path. Makes sense, will change it accordingly.

frankenjoe:
I changed to not raise an exception and extended the docstring with:

image

tests/test_io.py
```diff
@@ -610,14 +610,22 @@ def test_move_file(tmpdir, src_file, dst_file):
 @pytest.mark.parametrize(
     'path, new_extension, ext, expected_path',
     [
+        ('', '', None, ''),
+        ('', 'txt', None, ''),
```
hagenw, line 614: Could you add:

```suggestion
        ('', 'txt', None, ''),
        ('', '', 'rst', ''),
        ('', 'txt', 'rst', ''),
```

hagenw:
Great, thanks for adding these changes.

PR #91, opened 2023-01-27
Add support for short ID

### LINKED ISSUE #88: Add support for short UIDs
As discussed in https://github.com/audeering/audobject/issues/32 we should add support for short UIDs to `audeer`.

I would propose to solve this by:

* adding an `short=False` argument to `audeer.uid()`
* let `audeer.is_uid()` return `True` for short and long UIDs

As this is a breaking change for `audeer.is_uid()` we might consider doing a `2.0.0` release afterwards.

frankenjoe:
Closes #88 

In `audb` were we used for short IDs for the first time we started to use the last 8 digits, which we also do here. The 8 digits would have been a better choice as the following example shows:

```python
for short in False, True:
    for idx in range(2):
        uid = audeer.uid(short=short)
        print(uid)
```
```
db88e120-9e2f-11ed-a57a-80fa5b66caff
db88e121-9e2f-11ed-a57a-80fa5b66caff
5b66caff
5b66caff
```

This is also the reason I had to simplify the test for short IDs.

image

image

### DIFF
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -489,7 +489,7 @@ def is_uid(uid: str) -> bool:
         ``True`` if string is a unique identifier
 
     Examples:
-        >>> is_uid(uid())
+        >>> is_uid('626f68e6-d336-70b9-e753-ed9fad855840')
         True
         >>> is_uid(uid(short=True))
         True
--- a/audeer/core/utils.py
+++ b/audeer/core/utils.py
@@ -488,15 +488,34 @@ def is_uid(uid: str) -> bool:
     Returns:
         ``True`` if string is a unique identifier
 
+    Examples:
+        >>> is_uid(uid())
+        True
+        >>> is_uid(uid(short=True))
+        True
+        >>> is_uid('not a unique identifier')
+        False
+
     """
     if uid is None:
         return False
     if not isinstance(uid, str):
         return False
+    if len(uid) != 8 and len(uid) != 36:
+        return False
+
+    if len(uid) == 8:
+        uid = f'00000000-0000-0000-0000-0000{uid}'
+
+    for pos in [8, 13, 18, 23]:
+        if not uid[pos] == '-':
+            return False
+
     try:
         uuid.UUID(uid, version=1)
     except ValueError:
         return False
+
     return True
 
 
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -339,12 +339,18 @@ def test_is_semantic_version(version, is_semantic):
     'uid, expected',
     [
         (audeer.uid(), True),
+        (audeer.uid(short=True), True),
         (audeer.uid(from_string='from string'), True),
+        (audeer.uid(from_string='from string', short=True), True),
```
hagenw, line 344: I would also add a test for uppercase letters.

In https://en.wikipedia.org/wiki/Universally_unique_identifier they state:

> [RFC 4122 Section 3](https://tools.ietf.org/html/rfc4122#section-3) requires that the characters be generated in lower case, while being case-insensitive on input.

So, I guess we need something like:

```suggestion
        (audeer.uid(from_string='from string', short=True), True),
        (audeer.uid(from_string='from string', short=True).upper(), True),
```

audeer/core/utils.py
```diff
@@ -488,15 +488,34 @@ def is_uid(uid: str) -> bool:
...
         return False
+    if len(uid) != 8 and len(uid) != 36:
+        return False
+
+    if len(uid) == 8:
+        uid = f'00000000-0000-0000-0000-0000{uid}'
+
+    for pos in [8, 13, 18, 23]:
+        if not uid[pos] == '-':
+            return False
```
hagenw, line 512: Does this mean we had a bug before and were returning `True` for UIDs with `-` at the wrong position?

tests/test_utils.py
hagenw: But I have to admit I'm not completely sure about this.

From what we have written in the docstring, we would expect the test should fail as an uppercase UID cannot be created with `audeer.uid()`

audeer/core/utils.py
frankenjoe: Yes, it was returning `True` when I added a test where the `-` was in a wrong position.

tests/test_utils.py
frankenjoe: Ups. I accidentally committed your suggestion :)

frankenjoe: Where exactly do we say in the docstring that upper case is not supported?

hagenw: No, you are right. We only say they can be generated with `audeer.uid()`, but not that they have to.

hagenw: So I guess it is fine to stay with `True` for uppercase UIDs.

Issue #76, opened 2022-12-08
list_file_names(): raise error if path does not exist


frankenjoe:
We currently do not raise an error if `list_file_names()` is called with a non-existing path, see https://github.com/audeering/audbackend/pull/53#issuecomment-1342503423

For `list_dir_names()` we raise an error, but do not mention it in the docstring.

hagenw:
I think we do not raise an error with `audeer.list_file_names()` as path can be a folder, file or pattern. And as a pattern we recognize everything that is not an existing folder.

I think it will be very hard to change this behavior as we would need to find a solution that recognizes if the provided path is a pattern or not and only raise a `FileNotFoundError` if path is a folder or file. For a pattern it is reasonable to me to return an empty list if no files could be found.

My proposal would be to stay with the current behavior and not raise an error.

frankenjoe:
Mhh, but isn't it strange we raise an error with `audeer.list_dir_names('/does/not/exist')` but with `audeer.list_files_names('/does/not/exist')` we do not? Then we should maybe never raise an error and always return an empty list.

hagenw:
With `audeer.list_dir_names()`1 it makes sense to raise an error as we don't allow to provide a pattern to match path.

frankenjoe:
Or we raise an error:

1. if `path` ends on `/` and `path` does not exist
2. if `path` does not end on `/` and `os.path.dirname(path)` does not exist

?

hagenw:
As a solution to this issue seems not trivial, I would propose to postpone a possible solution to next year.

frankenjoe:
Sure, it's not urgent. We should also check what counts as a pattern, e.g. is `dir/to/path` a pattern or only if it includes some wildcards, e.g. `dir/to/path*` or `dir/to/*.path`.

frankenjoe:
Currently we have the following behavior:

```python
>>> audeer.mkdir('dir')
>>> audeer.touch('dir/file.txt')
>>> audeer.list_file_names('dir')
['dir/file.txt']
>>> audeer.list_file_names('d')
[]
>>> audeer.list_file_names('dir/file')
[]
>>> audeer.list_file_names('dir/file*')
['dir/file.txt']
```

And in my opinion the two cases where an empty list is returned should raise an error. At least that is also the behavior we expect with `ls`:

```bash
$ ls dir
file.txt
$ ls d
ls: cannot access 'd': No such file or directory
$ ls dir/file
[...]

PR #92, opened 2023-01-27
WIP: Allow path to be pattern in list_dir_names()

hagenw:


frankenjoe:
Wouldn't it make sense we first find a solution for https://github.com/audeering/audeer/issues/76? I guess with the changes you propose here we no longer raise an error if a directory does not exist, but return an empty list instead.

hagenw:
In my opinion #76 is solved by https://github.com/audeering/audeer/pull/93.

I guess it was not closed as I wrote "Closing" instead of "Closes" in the pull request ;)

frankenjoe:
As far as I see in #93 we only extended the docstring, but we didn't solve the actual issue, namely that the following returns an empty list:

```python
audeer.list_file_names('./does-not-exist/')
```
```
```

and I think the changes proposed introduce now the same issue for `list_dir_names()`.

frankenjoe:
Since we are struggeling to detect if `path` is a folder or a pattern in #94, I think it is not a good idea to introduce the same issue here. We should consider to introduce a separate `pattern` argument instead.

PR #94, opened 2023-01-30
list_file_names(): recognize if path is a pattern and raise error if folder does not exist

### LINKED ISSUE #76: list_file_names(): raise error if path does not exist

frankenjoe:
Closes #76 

We now distinguish the following three cases:

1. `path` is an existing file -> return `[file]`
2. `path` ends with a pattern (includes '*' or '?') -> raise error if `os.path.dirname(path)` does not exist, otherwise return a possibly empty list with matching files
3. otherwise we assume `path` is a directory and raise an error if it does not exist, otherwise a possibly empty list with matching files

image


frankenjoe:
In contrast to `ls` we return an empty list if `path` is a pattern that does not match files if `os.dirname(path)` exists. I think this behavior makes more sense.

hagenw:
image

This seems not like a good idea to me as we loose all the power of using real pattern then.
The following works with the current `main` branch:

```python
import audeer

audeer.mkdir('dir')
audeer.touch('dir/file1.wav')
audeer.touch('dir/file2.wav')
audeer.touch('dir/file3.wav')
audeer.list_file_names('dir/file[1-2].wav', basenames=True)
```
which returns
```
['file1.wav', 'file2.wav']
```
whereas with the current branch we get
```
NotADirectoryError: .../dir/file[1-2].wav
```

> In contrast to ls we return an empty list if path is a pattern that does not match files

I'm afraid that is exactly the problematic behavior. In order to achieve this we need to figure out if a pattern was given or not, but I would argue there is no way to figure this out.
This means we are left with the following options:
* raise an error if no file is matched by the pattern
* if `path` is not an existing file or folder we treat it as a pattern and only raise an error if `os.path.dirname(path)` does not exist and return an empty list if nothing is matched

audeer/core/io.py
```diff
@@ -474,11 +477,19 @@ def list_file_names(
 
     """
     path = safe_path(path)
-    if not os.path.isdir(path):
+
+    if os.path.isdir(path):
+        pattern = None
+    elif os.path.exists(path):
```
hagenw, line 483: If `path` exists and is not a directory it would be better to immediately return it (handling basenames first) and not call the helper function as this would need to first run `os.listdir()` which might be slow if we have a folder with millions of files.

frankenjoe: What should we do when `recursive` is to `True` in that case? E.g. if we have:

```
/root/file.txt
/root/sub/file.txt
```

and the user requests:

```
audeer.list_file_names('/root/file.txt`, recursive=True)
```

would you expect it returns only `/root/file.txt` or also `/root/sub/file.txt`. So far the behavior was to return also `/root/sub/file.txt`.

hagenw: I would expect that it should return only `'/root/file.txt'` if this matches a file as in this case.

hagenw: > a user wants to collect all paths with the same file name from all sub-folders

Yes, that's indeed a good use-case, but for that I would expect a requested path similar to `/root/*/file.txt`.
At least I don't understand why `'/root/file.txt'` should match a path that is `/root/sub/file.txt`.

frankenjoe: I guess you have to see the `file.txt` part as a sort of pattern just like with `file.*` but without wildcards. And if `/root/file.*` together with `recursive=True` returns all matching files from sub-folder, `file.txt` should do the same I think.

hagenw: To request `/root/*/file.txt` is of cause not possible at the moment, and I would also not support it.

One way to still get the result you are looking for is to force it to be a pattern instead of an actual existing file, e.g. `audeer.list_file_names("dir/*.txt", recursive=True)` or `audeer.list_file_names("dir/[f]ile.txt", recursive=True)` which is of cause also slightly strange ;)

hagenw: > you have to see the file.txt part as a sort of pattern just like with file.*

Maybe this is indeed the best option, and it is also the way it is implemented at the moment.
What we need to do then is to change the docstring from

```
path: path to file, directory or pattern
```
to something like
```
path: path to directory,
    or path to directory plus file matching pattern,
    e.g. ``'dir/file.txt'``.
    If ``recursive`` is ``True`` this would return all files named ``file.txt``
    from all sub-folders.
    Besides the filename `*`, `?`, `[seq]`, and `[!seq]` can be used
    as explained in :mod:`fnmatch`.
```

frankenjoe: > To request /root/*/file.txt is of cause not possible at the moment, and I would also not support it.

Right, that's what the `recursive` argument is for.

> which is of cause also slightly strange ;)

Yes, that's why I suggest we simply stick with the current behavior.

> audeer.list_file_names("dir/[f]ile.txt", recursive=True)

Hehe, good you mentioned that `[...]` is also a valid pattern. We currently do not recognize it as an pattern. Will change that.

frankenjoe: I changed to also recognize `[...]` as a pattern and extended the tests and docstring examples accordingly. To make it easier to write docstring examples the working directory is now set to a temporal directory, I.e. we no longer need to manually clean up temporal files afterwards.

audeer/core/io.py
```diff
@@ -664,7 +709,9 @@ def rmdir(
         NotADirectoryError: if path is not a directory
 
     Examples:
-        >>> mkdir('path1/path2/path3')  # doctest: +SKIP
```
frankenjoe, line 667: After changing to a temporal directory the docstring example failed because the folder `'path1/path2/path3'` did not exist. Took me a while to understand why, but I think `+SKIP` not only skips the output but also the whole expression, i.e. the directory was never created and the example only worked because the folder was created in a previous docstring example.

audeer/core/io.py
frankenjoe: I also changed the docstring of `path` to:

image

tests/test_io.py
```diff
@@ -268,6 +268,37 @@ def test_list_dir_names_errors(tmpdir):
...
+            ['file.txt'],
+            'file.txt',
+            '',
+            ['file.txt'],
+            False,
+            False,
+        ),
+        pytest.param(
+            [],
+            'file',
```
hagenw, line 282: Could you also add a test for `path='./file'`, I would expect this should return `[]` then.

hagenw: I think this part is still a little bit confusing.
In the docstring we state:

image

I would then interpret `file.txt` as a file matching pattern, otherwise you shouldn't also not be able to get more than one entry returned in the list.
The problem is when the user now presents just a single part of a path like `'file'` or `'t1.wav'` is this then interpreted as a path to a directory and we raise an error if the directory does not exist or should we interpret as a filename pattern?

audeer/core/io.py
```diff
@@ -420,14 +428,15 @@ def list_file_names(
         list of path(s) to file(s)
```
hagenw, line 428: Maybe we should add here that the list is sorted?

tests/test_io.py
hagenw: Your current solution seems to handle it like this:
if `path` does not exists as a file or directory and does not include any chars indicating it to be a pattern like `*`, `?`, ... it will interpret the path as a directory and not a pattern and raise an error as the folder does not exist.

If we stay with that solution, we need to update the docstring accordingly.

An alternative solution would be to treat the last part of the path (it's basename) always as search pattern as long as it don't end with `/` as you proposed in 1. in https://github.com/audeering/audeer/issues/76#issuecomment-1354558815.

tests/test_io.py
```diff
@@ -388,6 +424,93 @@ def test_list_dir_names_errors(tmpdir):
...
+            ],
+            '[ts]*.wav',
+            '',
+            ['s.wav', 's1.wav', 't1.wav', 't2.wav'],
+            False,
+            False,
+        ),
+        (
+            [],
+            os.path.join('file*'),
```
hagenw, line 482:
```suggestion
            'file*',
```

tests/test_io.py
```diff
@@ -388,6 +424,93 @@ def test_list_dir_names_errors(tmpdir):
...
+            '',
+            [],
+            False,
+            False,
+        ),
+        (
+            [],
+            os.path.join('?ile'),
```
hagenw, line 490:
```suggestion
            '?ile',
```

tests/test_io.py
```diff
@@ -268,6 +268,37 @@ def test_list_dir_names_errors(tmpdir):
         ([], '.', '', [], True, False),
         ([], '.', 'wav', [], False, False),
         ([], '.', 'wav', [], True, False),
+        # file
+        (
+            ['file.txt'],
+            'file.txt',
```
hagenw, line 274: Could you also add tests here where `file.txt` does only exist in a sub-folder and not in the main folder,
this should then return `['sub-folder/file.txt']`.

And a test for the case that `file1.txt` exists, but you ask for `file.txt`, which should raise an error with the current implementation.

hagenw: There is also a test missing for the case where we have a bunch of files, but the search pattern (including pattern chars like `*`) does not match any of them and we expect an empty list `[]`.

audeer/core/conftest.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def docdir(request):
+    # set temporal working directory in docstring examples
+    tmpdir = request.getfixturevalue('tmpdir')
+    with tmpdir.as_cwd():
```
hagenw, line 8: Cool, that's nice.
So far I always relied on cleaning up the root folder of the repo afterwards. This seems to be more elegant.

audeer/core/io.py
```diff
@@ -420,14 +428,15 @@ def list_file_names(
...
+            or ``path`` is a pattern
+            and ``os.dirname(path)`` does not exist
 
     Examples:
         >>> dir_path = mkdir('path')
         >>> _ = touch(os.path.join(dir_path, 'file.wav'))
+        >>> _ = touch(os.path.join(dir_path, 'File.wav'))
         >>> _ = touch(os.path.join(dir_path, '.lock'))
         >>> sub_dir_path = mkdir(os.path.join('path', 'sub'))
         >>> _ = touch(os.path.join(sub_dir_path, 'file.ogg'))
```
hagenw, line 442: In all those examples above we use currently `os.path.join()` as `path()` from `audeer` is not available as it is imported as `safe_path` so we can use the variable `path` in the source code.

If we want to promote `path()` here in the docstring examples we could maybe add it to the docstring namespace similar to what we do with `pd` in https://github.com/audeering/audmath/blob/main/audmath/core/conftest.py, but I'm not completely sure if this will make it too complicated to understand for new developers afterwards.

tests/test_io.py
frankenjoe: > An alternative solution would be to treat the last part of the path (it's basename) always as search pattern as long as it don't end with / as you proposed in 1

What is problematic with this solution is:

``` python
>>> audeer.path('dir/')
/dir
```

Of course, the better interface would have been to have separate arguments for the directory and an optional pattern. But now we are stuck with a single argument, so we need to figure out if `path` ends on a pattern or not.

tests/test_io.py
frankenjoe: > Could you also add tests here where file.txt does only exist in a sub-folder and not in the main folder,
this should then return ['sub-folder/file.txt'].

Well, with the current implementation, if `file.txt` does not exist then it's treated as folder and raises an error :) I added a test for it.

> There is also a test missing for the case where we have a bunch of files, but the search pattern (including pattern chars like *) does not match any of them and we expect an empty list [].

I added two more tests - one where we only match files in a sub-directory, and one where we do not match any file.

tests/test_io.py
frankenjoe: changed locally

tests/test_io.py
frankenjoe: changed locally

audeer/core/io.py
frankenjoe: If we think this is important, I suggest we do it in a separate PR as probably also other examples are affected. Another solution would be to simplify the example and to create the files in the fixture instead.

tests/test_io.py
```diff
@@ -277,6 +277,15 @@ def test_list_dir_names_errors(tmpdir):
...
             False,
         ),
+        pytest.param(
+            [os.path.join('sub', 'file.txt')],
+            'file.txt',
+            '',
+            [],
+            False,
+            False,
+            marks=pytest.mark.xfail(raises=NotADirectoryError),
```
hagenw, line 287: Can you repeat this test with `recursive=True` there it should not fail, but return the file pattern.

tests/test_io.py
hagenw: Yes, we need to stick with the single argument. When we don't want to use `/` at the end, which also seems risky to me, there are the following alternatives:

* every time path includes a `/` the second part is treated as a pattern, which will have the side effect that `'file'` and `'./file'` might return different results
* implement it as `ls` and raise also an error if no file is matched by the pattern
* only raise an error if `os.path.dirname(path)` does not exist and return empty list in all other none-matched, none-existing cases

frankenjoe: > every time path includes a / the second part is treated as a pattern, which will have the side effect that 'file' and './file' might return different results

If I understand you correctly, this would mean we generally remove support for listing all files in a folder, which seems to me the most common use-case of the function.

> implement it as ls and raise also an error if no file is matched by the pattern

This would mean we get an empty list for `list_file_names('/empty/folder')` but an error with `list_file_names(''/empty/folder/*')`.

> only raise an error if `os.path.dirname(path)` does not exist and return empty list in all other none-matched, none-existing cases

Not sure, but I think this is what we already do at the moment. So we would not solve #67.

Actually I have a fourth suggestion and that is we leave `list_file_names()` as is, but deprecate it in favor of a new function where we separate folder and pattern, e.g. `list_files()`. We can then also get rid of `filetype`, which is not really needed since a user can do `*.<ext>` and if we want to change the default parameter of `hidden` to `True` as we do with `list_dir_names()`.

tests/test_io.py
frankenjoe: As I pointed out before, it will still fail. Since `file.txt` does not exist and it is not pattern, it will treat it as directory and since the directory does not exist, it raises an error.

audeer/core/io.py
hagenw: I created https://github.com/audeering/audeer/issues/95 to track this.

tests/test_io.py
hagenw: I was also expecting it to fail, but I would argue that is not the behavior a user would expect.
Because now what `list_file_names()` returns for the sub folder does not depend on the actual content of the sub-folder, but on the content of the main folder, which feels very strange.

tests/test_io.py
frankenjoe: > Having an additional pattern variable feels also a little bit strange as ls just works in bash.

Yes, but also only with the kind of strange side-effect that it raises an error if the directory exists but no files match, which I don't like.

Otherwise, I'm also fine with the solution proposed here. The only downside I see is that `list_file_names('./file.txt', recursive=True)` will raise an error if `./file.txt` does not exist but only `./sub/file.txt`. In that case I would expect it returns `['./sub/file.txt']`, but since `./file.txt` is not recognizable as a pattern it will treat it as a directory and raise an error.

tests/test_io.py
frankenjoe: I know, but I don't see a solution to this as long as we don't introduce a separate `pattern` argument.

So maybe we should only support `recursive=True` when `path` is a folder or ends on a pattern. I.e.  `list_file_names('./file.txt', recursive=True)` will either fail if `'./file.txt'` does not exist or return `['./file.txt']` (even if e.g. `'./sub/file.txt'` exists).

audeer/core/io.py
```diff
@@ -410,7 +410,15 @@ def list_file_names(
     """List of file names inferred from provided path.
 
     Args:
-        path: path to file, directory or pattern
+        path: path to directory,
+            or path to directory plus file matching pattern,
+            e.g. ``'dir/file.txt'``.
+            If ``recursive`` is ``True``
```
hagenw, line 416: To adjust it to the behavior we introduce here:

```suggestion
            If ``file.txt`` exists in ``dir``
            and ``recursive`` is ``True``
```

audeer/core/io.py
```diff
@@ -420,14 +428,15 @@ def list_file_names(
         list of path(s) to file(s)
 
     Raises:
-        NotADirectoryError: if ``os.path.dirname(path)``
-            is not a directory
-        FileNotFoundError: if ``os.path.dirname(path)``
-            does not exists
+        NotADirectoryError: if ``path`` is a directory
+            that does not exist,
```
hagenw, line 432: It will also raise the error if `path` is not a pattern, but a file that does not exist.
So maybe change to
```suggestion
        NotADirectoryError: if ``path`` does not exist,
```

or to 

```suggestion
        NotADirectoryError: if ``path`` is a directory or file
            that does not exist,
```

tests/test_io.py
hagenw: I proposed changes to the docstring to handle the `file.txt` does not exists case, and agree with you that we should still include the changes to `audeer.list_file_names()`.

We can then still decide if we need an additional `audeer.list_files()` solution.

tests/test_io.py
hagenw: There might also be the possibility to implement a solution to return `file.txt` from a sub-folder.
E.g. we could not only check `os.path.exists(path)`, but we could do it for every sub-folder if `recursive` is `True`.

frankenjoe: That's indeed a solution. I implemented your suggestion and now the following works as expected:

```python
>>> audeer.mkdir('dir')
>>> audeer.mkdir('dir/sub')
>>> audeer.touch('dir/sub/file.txt')
>>> audeer.list_file_names('file.txt', basenames=True, recursive=True))
['dir/sub/file.txt']
```

There's one more special case that we now treat as follows:

```python
>>> audeer.touch('.file.txt')
>>> audeer.list_file_names(.file.txt', basenames=True)
[]
>>> audeer.list_file_names(.file.txt', basenames=True, hidden=True)
['.file.txt']
```

Or would you expect it raises an error in the first case?

tests/test_io.py
frankenjoe: We now have a solution for the missing case, see https://github.com/audeering/audeer/pull/94#discussion_r1098763187, so I guess there is not really a need for introducing a new function.

audeer/core/io.py
frankenjoe: We found a solution to return `file.txt` from sub-folders even if it does not exist in the root folder.

audeer/core/io.py
frankenjoe: Changed to:

image

tests/test_io.py
hagenw: No, I think this is a good solution.

It also works for a combination of hidden files in sub-folders:

```python
>>> audeer.mkdir(sub')
>>> audeer.touch('sub/.file.txt')
>>> audeer.list_file_names('.file.txt', basenames=True, recursive=True, hidden=True)
['sub/.file.txt']
>>> audeer.list_file_names('.file.txt', basenames=True, hidden=True)
...
NotADirectoryError: /home/audeering.local/hwierstorf/git/audeering/audeer/.file.txt
```

But there is one case where it is not working yet:

```python
>>> # the next should return []
>>> audeer.list_file_names('.file.txt', basenames=True, recursive=True)
...
NotADirectoryError: /home/audeering.local/hwierstorf/git/audeering/audeer/.file.txt
```

frankenjoe: Good catch. This case is now covered and I added a test for it.

hagenw: Cool, thanks.

hagenw:
To summarize the changes compared to the current implementation I executed the new tests in the `main` branch and found the following differences:

| Existing files | Executed code | New behavior | Old behavior |
| --- | ---  |--- | --- |
|`['sub/file.txt']` | `list_file_names('file.txt')` | `NotADirectoryError` | `[]` |
| `[]` | `list_file_names('file')` | `NotADirectoryError` | `[]` |
| `[]` | `list_file_names('does-not-exist')` | `NotADirectoryError` | `[]` |
| `[]` | `list_file_names('does/not/exist')` | `NotADirectoryError` | `FileNotFoundError` |
| `[]` | `list_file_names('does/not/exist/file*')` | `NotADirectoryError` | `FileNotFoundError` |
| `[]` | `list_file_names('not!a[pattern')` | `NotADirectoryError` | `[]` |

To me, this all looks good. Does it also align with your expectations?

frankenjoe:
> To me, this all looks good. Does it also align with your expectations?

yes

PR #97, opened 2023-04-14
test Windows

hagenw:
Just to see what `audeer.path('a/b')` return under Windows `/` or `\`


hagenw:
`audeer.path('a/b')` does change `/` to `\`, which it has to do anyway, because it also has to convert it to an absolute path which should not contain a mixture of `/` and `\`.

Issue #102, opened 2023-04-27
create_archive(): allow files to be None


frankenjoe:
It might be convenient to support `files=None` with `create_archive()` and in that case include all files below `root`.

hagenw:
Added with version 1.20.0.

PR #103, opened 2023-04-27
Fix archive errors and support files is None

### LINKED ISSUE #101: create_archive(): enforce files exist within root?
In the docstring of `create_archive()` we state:

image

But there are no checks that the files are relative paths and exist below `root`. 

This can have probably undesired side-effects, e.g.:

```python
with tempfile.TemporaryDirectory() as tmp:
    file = audeer.touch(f'{tmp}/a')
    audeer.create_archive('.', file, 'a.zip')
```

This includes a file from a temporary folder outside the working directory and when the archive is unpacked creates it within the working directory:

```python
files = audeer.extract_archive(
    'a.zip',
    '.',
)
print(files)
```
```
['tmp/tmpx7hnjy3m/a']
```

I wonder if we need checks that ensure that `files` are relative paths (also not leading outside the root directory) or at least exist within root.

frankenjoe:
Closes #101 #102 

* Check if files are below root
* Check that files exist
* Check that archive is not a folder
* Check that root / destination is a folder
* Extend raises sections
* Include all files if files is None

image

image

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -98,7 +98,7 @@ def common_directory(
 
 def create_archive(
         root: str,
-        files: typing.Union[str, typing.Sequence[str]],
+        files: typing.Optional[typing.Union[str, typing.Sequence[str]]],
         archive: str,
         *,
         verbose: bool = False,
@@ -111,7 +111,10 @@ def create_archive(
             will be relative to ``root``
         files: files to include in archive.
             Path can be absolute or relative
-            but must be below ``root``
+            but must be below ``root``.
+            If set to ``None``
+            will include all files below ``root``
+            (also from sub-folders and hidden files)
         archive: path to archive file.
             The archive type is determined by the file extension
         verbose: if ``True`` a progress bar is shown
@@ -126,7 +129,6 @@ def create_archive(
     root = safe_path(root)
     archive = safe_path(archive)
     mkdir(os.path.dirname(archive))
-    files_org = to_list(files)
 
     if not os.path.exists(root):
         raise FileNotFoundError(
@@ -142,36 +144,48 @@ def create_archive(
             root,
         )
 
-    files = []
+    if files is None:
 
-    for file in files_org:
+        files = list_file_names(
+            root,
+            basenames=True,
+            recursive=True,
+            hidden=True,
+        )
 
-        # convert to absolute path
-        if not os.path.isabs(file):
-            path = safe_path(root, file)
-        else:
-            path = safe_path(file)
+    else:
 
-        # file is below root
-        if not path.startswith(root):
-            raise ValueError(
-                f"Only files below "
-                f"'{root}' "
-                f"can be included. "
-                f"This is not the case with "
-                f"'{file}'"
-            )
+        files_org = to_list(files)
+        files = []
 
-        # file exists
-        if not os.path.exists(path):
-            raise FileNotFoundError(
-                errno.ENOENT,
-                os.strerror(errno.ENOENT),
-                file,
-            )
+        for file in files_org:
 
-        # convert to relative path
-        files.append(path[len(root) + 1:])
+            # convert to absolute path
+            if not os.path.isabs(file):
+                path = safe_path(root, file)
+            else:
+                path = safe_path(file)
+
+            # file is below root
+            if not path.startswith(root):
+                raise ValueError(
+                    f"Only files below "
+                    f"'{root}' "
+                    f"can be included. "
+                    f"This is not the case with "
+                    f"'{file}'"
+                )
+
+            # file exists
+            if not os.path.exists(path):
+                raise FileNotFoundError(
+                    errno.ENOENT,
+                    os.strerror(errno.ENOENT),
+                    file,
+                )
+
+            # convert to relative path
+            files.append(path[len(root) + 1:])
 
     # Progress bar arguments
     desc = format_display_message(
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -126,7 +126,7 @@ def create_archive(
         FileNotFoundError: if ``root`` or a file in ``files`` is not found
         NotADirectoryError: if ``root`` is not a directory
         RuntimeError: if archive does not end with ``zip`` or ``tar.gz``
-        ValueError: if a file in ``files`` is not below ``root``
+            or a file in ``files`` is not below ``root``
 
     """
     root = safe_path(root)
@@ -171,7 +171,7 @@ def create_archive(
 
             # file is below root
             if not path.startswith(root):
-                raise ValueError(
+                raise RuntimeError(
                     f"Only files below "
                     f"'{root}' "
                     f"can be included. "
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -126,7 +126,7 @@ def create_archive(
         FileNotFoundError: if ``root`` or a file in ``files`` is not found
         NotADirectoryError: if ``root`` is not a directory
         RuntimeError: if archive does not end with ``zip`` or ``tar.gz``
-        ValueError: if ``file`` is not below ``root``
+        ValueError: if a file in ``files`` is not below ``root``
 
     """
     root = safe_path(root)
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -105,6 +105,16 @@ def create_archive(
 ):
     r"""Create ZIP or TAR.GZ archive.
 
+    If a list with ``files`` is provided,
+    only those files will be included to the archive.
+    In that case the files are added in the given order,
+    which may have an influence on the checksum of the archive.
+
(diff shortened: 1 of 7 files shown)


audeer/core/io.py
```diff
@@ -106,21 +107,88 @@ def create_archive(
...
+
+            # convert to absolute path
+            if not os.path.isabs(file):
+                path = safe_path(root, file)
+            else:
+                path = safe_path(file)
+
+            # file is below root
+            if not path.startswith(root):
+                raise ValueError(
```
frankenjoe, line 174: If archive does not end on ZIP or TAR.GZ we raise a `RuntimeError`. Should we raise a `RuntimeError` for consistency here, too?

hagenw: Yes, I would change to `RuntimeError`.

audeer/core/io.py
```diff
@@ -278,17 +373,25 @@ def extract_archives(
...
-            Will be created if it doesn't exist
+        destination: folder where the files will be extracted.
+            If the folder does not exists,
+            it will be created
         keep_archive: if ``False`` delete archive files after extraction
         verbose: if ``True`` a progress bar is shown
 
     Returns:
-        combined member filenames of archives
+        extracted members
```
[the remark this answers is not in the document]
hagenw: Yes, makes sense to me. Maybe you can just replace `m.name` by `os.path.normpath(m.name)`.

audeer/core/io.py
```diff
@@ -106,21 +107,88 @@ def create_archive(
...
-            The archive type is determined by the file extension
+            The type of the archive
+            is determined from its file extension
         verbose: if ``True`` a progress bar is shown
 
     Raises:
+        FileNotFoundError: if ``root`` or a file in ``files`` is not found
+        NotADirectoryError: if ``root`` is not a directory
         RuntimeError: if archive does not end with ``zip`` or ``tar.gz``
+        ValueError: if ``file`` is not below ``root``
```
hagenw, line 129:
```suggestion
        ValueError: if a file in ``files`` is not below ``root``
```

tests/test_io.py
```diff
@@ -8,86 +8,305 @@
...
+        ),
+        (  # root is sub folder
+            ['sub/file.txt'],
+            './sub',
+            ['file.txt'],
+            'archive.zip',
+            'archive.zip',
+            '.',
+            ['file.txt'],
+        ),
```
hagenw, line 127: I would repeat this test, but don't provide `files` to see if it is able to figure out the correct names without including `root`:

```suggestion
        (  # root is sub folder
            ['sub/file.txt'],
            './sub',
            ['file.txt'],
            'archive.zip',
            'archive.zip',
            '.',
            ['file.txt'],
        ),
        (
            ['sub/file.txt'],
            './sub',
            None,
            'archive.zip',
            'archive.zip',
            '.',
            ['file.txt'],
        ),
```

tests/test_io.py
```diff
@@ -8,86 +8,305 @@
...
+            ['.hidden', 'file.txt', 'sub/a/b/file.txt'],
+        ),
+        (  # tar.gz
+            ['file.txt', 'sub/a/b/file.txt'],
+            '.',
+            ['sub/a/b/file.txt', 'file.txt'],
+            'archive.tar.gz',
+            'archive.tar.gz',
+            '.',
+            ['sub/a/b/file.txt', 'file.txt'],
```
hagenw, line 117: Is the order of the output given by the order of files, or are folders always listed before files?

tests/test_io.py
```diff
@@ -8,86 +8,305 @@
...
+            ['file.txt'],
+            '.',
+            [],
+            'archive.zip',
+            'archive.zip',
+            'file.txt',
+            [],
+            marks=pytest.mark.xfail(raises=NotADirectoryError),
+        ),
+        pytest.param(  # extracted archive does not exit
```
hagenw, line 176:
```suggestion
        pytest.param(  # archive to be extracted does not exit
```

tests/test_io.py
```diff
@@ -8,86 +8,305 @@
...
+            [],
+            '.',
+            [],
+            'archive.zip',
+            'bad.zip',
+            '.',
+            None,
+            marks=pytest.mark.xfail(raises=FileNotFoundError),
+        ),
+        pytest.param(  # extracted archive is a directory
```
hagenw, line 186:
```suggestion
        pytest.param(  # archive to be extracted is a directory
```

tests/test_io.py
```diff
@@ -8,86 +8,305 @@
...
+        assert os.path.exists(archive_extract)
+
+    # delete archive
+
+    audeer.extract_archive(
+        archive_extract,
+        destination,
+        keep_archive=False,
+    )
+    assert not os.path.exists(archive_extract)
```
hagenw, line 309: Maybe also add a test that the extracted files exist at the expected place.

audeer/core/io.py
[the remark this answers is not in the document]
hagenw: Maybe it's also worse adding an `if`-clause to change the path only when we running Windows as it might slow the process down if you have millions of files in the archive.

tests/test_io.py
frankenjoe: Yes, that is very unfortunate. But it's the order in which the files are provided to the zip files. The proper solution would have been to sort the files before passing them to the zip file cause now the checksum of the archive depends on the order in which files are passed, e.g.:

``` python
with open('a', 'w') as fp:
    fp.write('a')

with open('b', 'w') as fp:
    fp.write('b')

audeer.create_archive('.', ['a', 'b'], 'ab.zip')
print(audbackend.md5('ab.zip'))

audeer.create_archive('.', ['b', 'a'], 'ba.zip')
print(audbackend.md5('ba.zip'))
```
```
2744ad8eda2f408ac4a84a8a765e4e3a
2505d3354e4c2813aa260839c0f99a60
```

At least with `files=None` we now make sure that files are sorted.

If we want we could at least sort the result of `extract_archive()` if we want to align it with our functions that usually always return files in sorted order.

hagenw: Interestingly, you could also call it a feature.

I would propose the following:
* sort files for `files=None`
* in the docstring state under `files` that the order of the provided files will influence the MD5 sum of the archive and that for `files=None` we sort the files

Regarding sorting at the end I'm not completely sure as it will slow down the process and it might not be needed by every user.

audeer/core/io.py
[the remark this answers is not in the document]
frankenjoe: > Maybe you can just replace m.name by os.path.normpath(m.name).

I think it's enough to do `m.name.replace('/', os.path.sep)` as we call `audeer.path()` on the files before adding them to the zip and that includes `normpath()`, right?

> Maybe it's also worth adding an if-clause to change the path only when we running Windows as it might slow the process down if you have millions of files in the archive.

Good idea.

tests/test_io.py
frankenjoe: Maybe:

image

and:

image

audeer/core/io.py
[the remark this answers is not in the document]
hagenw: >> Maybe you can just replace m.name by os.path.normpath(m.name).
>
>I think it's enough to do m.name.replace('/', os.path.sep) as we call audeer.path() on the files before adding them >to the zip and that includes normpath(), right?

Yes, `m.name.replace('/', os.path.sep)` is sufficient.

tests/test_io.py
hagenw: Yes, looks good

tests/test_io.py
frankenjoe: done

tests/test_io.py
frankenjoe: done

audeer/core/io.py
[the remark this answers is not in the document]
frankenjoe: done, and changed to:

image

audeer/core/io.py
frankenjoe: done

hagenw:
> @hagenw I just realized we are missing docstring examples. Should I add some in this PR or do you prefer we do it in another PR as we already have so many changes by now.

Let's do it in another PR.

tests/test_io.py
frankenjoe: I changed to:

```python
    result = audeer.extract_archive(
        archive_extract,
        destination,
        keep_archive=False,
    )
    assert result == expected
    assert not os.path.exists(archive_extract)
```

but I guess that was not what you were asking for then :)

frankenjoe: Do you mean something like:

```
    assert result == expected
    for file in result:
        assert os.path.exists(audeer.path(root, file))
```

?

hagenw: Yes, exactly.

PR #108, opened 2023-05-02
TST: exclude Windows only code from coverage

### LINKED ISSUE #107: Coverage can no longer be reached locally
In https://github.com/audeering/audeer/pull/103 we introduced to use platform specific coverage configurations.

The problem is that when executing the tests locally as proposed in `CONTRIBUTING.rst`:

```bash
$ python -m pytest
```

the coverage can no longer be reached on Linux due to https://github.com/audeering/audeer/blob/6fc395486d5fe1ce18a6701485c1f53c398a9a05/audeer/core/io.py#L375

I see three possible solutions to the problem:

1. Update `CONTRIBUTING.rst` and state that you need to run `python -m pytest --cov-config=.coveragerc.<platform>`
2. Replace `pragma: no Linux or macOS cover` by `pragma: no cover` ~~as we do not test coverage under Windows anyway~~
3. Use [tox](https://tox.wiki/en/latest/) to manage tests as it [supports coverage on different platforms](https://www.topbug.net/blog/2020/12/19/platform-dependent-python-coverage-test-with-tox/)

Solution 1 is easy to do, but it is very cumbersome that you cannot just run `python -m pytest`.
I would be in favor of solution 2.

hagenw:
Closes #107 

Exclude Windows only code from coverage to be able to reach full coverage when just running
```bash
$ python -m pytest
```

### DIFF
--- a/.coveragerc.Linux
+++ b/.coveragerc.Linux
@@ -2,4 +2,3 @@
 exclude_lines =
     pragma: no cover
     pragma: no Linux cover
-    pragma: no Linux or macOS cover
--- a/.coveragerc.Linux
+++ b/.coveragerc.Linux
@@ -1,4 +1,0 @@
-[report]
-exclude_lines =
-    pragma: no cover
-    pragma: no Linux cover
--- a/.coveragerc.Windows
+++ b/.coveragerc.Windows
@@ -1,4 +1,0 @@
-[report]
-exclude_lines =
-    pragma: no cover
-    pragma: no Windows cover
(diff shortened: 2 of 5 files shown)


frankenjoe:
I wonder if it would still make sense to keep the platform dependent cover files. Maybe in the future we will need them at some point?

hagenw:
> I wonder if it would still make sense to keep the platform dependent cover files. Maybe in the future we will need them at some point?

The problem is that we will always have the problem then that the coverage might not be reached locally.
The only way to really handle different coverage for different platforms would be by solution 1 or 3 from #107, but not by the solution I propose here.

frankenjoe:
Ok, then let's simply remove those coverage files for now and only if we run into the issue that we need to execute different code on all platforms we can re-introduce them or fine a different solution.

PR #126, opened 2023-11-30
Support *paths in mkdir() and rmdir()

### LINKED ISSUE #64: Extend path related functions to audeer.path() behavior
In `audeer.path()` we can now provide several arguments that are then joined by `os.path.join()`.
Would be great to extend `audeer.mkdir()`, `audeer.rmdir()` to behave the same.

hagenw:
Closes #64 

This extends `audeer.mkdir()` and `audeer.rmdir()` to support several path arguments, so we can replace
```python
audeer.mkdir(audeer.path('a', 'b'))
audeer.rmdir(audeer.path('a', 'b'))
```
with
```python
audeer.mkdir('a', 'b')
audeer.rmdir('a', 'b')
```

I realized that I'm typing very often `audeer.mkdir(audeer.path(...))` in my daily coding,
so the above changes would make it easier.
The downside is that `audeer.rmdir('a', 'b')` might be ambiguous for a user to grasp what it is doing, e.g. deleting `a/b`, or deleting `a` and `b`. 
@audeerington any opinion if we should change it or not?

---

image

---

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -499,7 +499,7 @@ def list_dir_names(
         FileNotFoundError: if path does not exists
 
     Examples:
-        >>> _ = mkdir('path/a/.b/c')
+        >>> _ = mkdir('path', 'a', '.b', 'c')
         >>> list_dir_names(
         ...     'path',
         ...     basenames=True,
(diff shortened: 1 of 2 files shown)


frankenjoe:
> I realized that I'm typing very often audeer.mkdir(audeer.path(...)) in my daily coding,

Me, too :) I like the changes, though I rarely use `rmdir()`, but I think it makes sense to support it there, too.

PR #127, opened 2023-12-05
Support *paths in touch()

hagenw:
Similar to https://github.com/audeering/audeer/pull/126 this adds support for `*paths` to `audeer.touch()`.
This allows a syntax like `audeer.touch('folder', 'file.txt')`.

We have a few more functions that have `path` as input, but we cannot add support for `*paths` to all of them, e.g. `move_file()` requires `src_path` and `dst_path`. Besides `audeer.touch()`, we might consider to add it to `audeer.list_file_names()` and `audeer.list_dir_names()`, but I'm not completely convinced ;)

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -576,12 +576,12 @@ def list_file_names(
 
     Examples:
         >>> dir_path = mkdir('path')
-        >>> _ = touch(os.path.join(dir_path, 'file.wav'))
-        >>> _ = touch(os.path.join(dir_path, 'File.wav'))
-        >>> _ = touch(os.path.join(dir_path, '.lock'))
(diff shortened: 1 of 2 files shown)


audeer/core/io.py
audeerington, line 1: The examples in the docstring of `list_file_names` and `move_file` could be updated to use this new feature.

audeerington (commented):
I agree, I think adding `*paths` for `touch` makes sense, but would also tend not to add it for the other functions.

audeer/core/io.py
hagenw: Good idea, I updated both docstrings.

audeerington: Great, everything looks good to me then.

PR #135, opened 2024-01-18
Do now follow symlinks in download_url()

hagenw:
Based on https://github.com/audeering/audeer/pull/132 this changes `audeer.download_url()` to call `audeer.path(..., follow_symlink=False)` to make it faster when used in a loop. It does not have an effect on its behavior, but introduces a breaking change as it now returns a symlink and not the realpath if `destination` was given as a symlink.

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -256,7 +256,7 @@ def download_url(
         'favicon.png'
 
     """  # noqa: E501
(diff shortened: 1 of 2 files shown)


hagenw:
This is no longer needed as it is the new default, see https://github.com/audeering/audeer/pull/140

PR #137, opened 2024-01-18
TST: test for symbolic link in audeer.rmdir()

hagenw:
This adds a test for `audeer.rmdir()` which fails when changing to `audeer.path(..., follow_symlink=False)` inside `audeer.rmdir()`.


hagenw:
Closing in favor of #141

PR #138, opened 2024-01-18
Don't convert symlinks in audeer.extract_archive()

hagenw:
Speeds up `audeer.extract_archive()` by using `audeer.path(..., follow_symlink=False)`.


hagenw:
This is no longer needed as it is the new default, see https://github.com/audeering/audeer/pull/140

Issue #136, opened 2024-01-18
Speed up functions that use audeer.path()


hagenw:
In https://github.com/audeering/audeer/pull/132 we introduced the `follow_symlink` argumet to `audeer.path()` which does not convert symlinks into the actual links, which makes it much faster.

I checked all functions, that use `audeer.path()` and add a discussion here, how we might handle applying the changes there too:

* `audeer.basename_wo_ext()`: I would silently use `follow_symlink=False` as the current behavior for symlinks feels like a bug for me (the returned name might be different if the basename of the link is different from the file name) https://github.com/audeering/audeer/pull/139
* `audeer.common_directory()`: I'm not sure what the wanted behavior is here. At the moment it would first convert symlink, which means `a/b/c` and `d/e/f` can have a common directory if at least some it is a link. If this is the desired behavior, we should not change it or add `follow_symlink` as an argument.
* `audeer.create_archive()`: it is possible to silently use `follow_symlink=False`. It would introduce a breaking change as a user can then no longer include symlinks in the `files` argument that are outside of `root`, but just link into sub-folders of `root`.
* `audeer.download_url()` we can silently use `follow_symlink=False`. The only breaking change is that it then would return a symlink instead of the real path (https://github.com/audeering/audeer/pull/135)
* `audeer.extract_archive()`: we should be able to silently use `follow_symlink=False` without any breaking change (https://github.com/audeering/audeer/pull/138)
* `audeer.extract_archives()`: we should be able to silently use `follow_symlink=False` without any breaking change (https://github.com/audeering/audeer/pull/138)
* `audeer.file_extension()`: we need to add a `follow_symlink` argument and pass it on to `audeer.path()` as it leads to different outputs. Or we could argue as with `audeer.basename_wo_ext()` that the current behavior is more a bug than a feature and change to use `follow_symlink=False`
* `audeer.list_dir_names()`: I would add a `follow_symlink` option as the output is directly affected or do nothing
* `audeer.list_file_names()`: I would add a `follow_symlink` option as the output is directly affected or do nothing
* `audeer.mkdir()`: we could silently use `follow_symlink=False`, but it would change the return value. Maybe we can skip it as speed is anyway not an issue here as we have to do an hard disc operation anyway
* `audeer.md5()`: we would need to add `follow_symlink=False` as an option as it directly influences the return value. If we add it here, we also need to add it at `audeer.list_file_names()`
[...]

PR #139, opened 2024-01-18
Not follow symlinks in Ensure basename_wo_ext()

hagenw:
Relates to #136 

The result of `audeer.basename_wo_ext()` should not be influenced by existing files or symlinks.

This adds first a failing test for the current behavior and then a fix, that avoids following symlinks.


hagenw:
This is now longer needed as this is the new default, see https://github.com/audeering/audeer/pull/140

PR #140, opened 2024-01-19
follow_symlink=False as default in path()

hagenw:
Relates to https://github.com/audeering/audeer/issues/136

Changes the default of `follow_symlink` in `audeer.path()` to `False`.

Also fixes handing the `follow_symlink` settings to `audeer.path()` in `audeer.safe_path()`.

### DIFF
--- a/audeer/core/path.py
+++ b/audeer/core/path.py
@@ -17,7 +17,7 @@ if platform.system() in ["Darwin", "Windows"]:  # pragma: no cover
 def path(
     path: typing.Union[str, bytes],
     *paths: typing.Sequence[typing.Union[str, bytes]],
-    follow_symlink: bool = True,
+    follow_symlink: bool = False,
 ) -> str:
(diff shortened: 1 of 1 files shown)


hagenw:
All tests are passing as we don't test for symbolic links in most of them.
Looking at #136 the most critical functions seem to be:
* `audeer.md5()` as this pull request will change the MD5 sum when a symbolic link is given
* `audeer.rmdir()` as this will now fail for a symbolic link instead of silentely removing it and the original file

frankenjoe:
To me it makes sense we call `audeer.path()` with `follow_symlink=True` in `audeer.md5()` and `audeer.rmdir()` since we are not interested in the MD5 of the symbolic link or calling remove directory on it.

hagenw:
OK, to better keep track of the changes, I will first merge this one and then take care of `audeer.rmdir()` and `audeer.md5()` in seperate pull requests.

PR #142, opened 2024-01-22
TST: symlink in md5() and list_file_names()

hagenw:
This adds tests for the handling of symbolic links to `audeer.md5()` and `audeer.list_file_names()` (as this is used inside `audeer.md5()`). 

It turns out that all those newly added tests do also pass when we change the default of `follow_symlink` to `True` in `audeer.path()` as we did already not convert symlinks in `audeer.list_file_names()`, which is used inside `audeer.md5()` to list file names in a folder. This means we do not need to adapt `audeer.md5()` at all for https://github.com/audeering/audeer/pull/140.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -985,7 +985,11 @@ def test_list_file_names_symlinks(tmpdir):
     files = audeer.list_file_names(folder, basenames=True)
     assert files == ["file.txt"]
     files = audeer.list_file_names(folder, basenames=True, recursive=True)
-    assert files == ["file.txt", "link/file2.txt", "sub-folder/file2.txt"]
+    assert files == [
+        "file.txt",
+        os.path.join("link", "file2.txt"),
+        os.path.join("sub-folder", "file2.txt"),
+    ]
 
 
 def test_md5_errors():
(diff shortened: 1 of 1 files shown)


frankenjoe:
> It turns out that all those newly added tests do also pass when we change the default of follow_symlink to True in audeer.path() 

Do you mean "...to False"?

> This means we do not need to adapt audeer.md5() at all for https://github.com/audeering/audeer/pull/140.

Does it mean that we do not return the checksum of the file that is linked, but of the link itself?

hagenw:
> > It turns out that all those newly added tests do also pass when we change the default of follow_symlink to True in audeer.path()
> 
> Do you mean "...to False"?

No, `False` is the new default. I mean when I locally change the default of `follow_symlink` in `audeer.path()` back to `True` to mimick its old behavior the new tests do still pass.
 
> > This means we do not need to adapt audeer.md5() at all for #140.
> 
> Does it mean that we do not return the checksum of the file that is linked, but of the link itself?

If you request the checksum of a file, it will always return the checksum based on the files content, independent if the file is a symlink or not.
But were the name of the symlink/file comes into play is if we calculate the checksum of a folder, there we create the checksum based on the content of all files in that folder, but also based on the **name** of the files. And for the name we did not converted a symlink to the actual file name before. Hence, there is no need to change anything as the behavior is not affected by `audeer.path()`.

frankenjoe:
Ok, great. Thanks for the clarifications.

PR #147, opened 2024-07-01
Add maximum_refresh_time to progress_bar()

hagenw:
This adds `maximum_refresh_time` which allows to ensure that a progress bar is refreshed at least every `maximum_refresh_time` seconds.

This is of interest when the time between the single steps of the progress bar iteration take very long, and a user might start to wonder if the process is stalled or still running.

The refreshing is achieved by using another thread as proposed in https://github.com/tqdm/tqdm/issues/861#issuecomment-2197893883. To not change and possible break any existing code, the default value of `maximum_refresh_time` is `None` which will not force refreshing the progress bar. I would propose we set this value then to an actual value in applications where it is relevant, e.g. `audb`.

For a local example run:

```python
import time

import audeer


for _ in audeer.progress_bar(range(2)):  # old behavior
    time.sleep(4)
```

which results in a display of `00:00` for a long time:

image

and compare with
```python
for _ in audeer.progress_bar(range(2), maximum_refresh_time=1):
    time.sleep(2)
```

which updates the already past time every second:

image

Updated docstring:

image

### DIFF
--- a/tests/test_tqdm.py
+++ b/tests/test_tqdm.py
@@ -37,3 +37,15 @@ def test_progress_bar():
     pbar = audeer.progress_bar([0.1])
     for step in pbar:
         time.sleep(step)
+
+
+def test_progress_bar_update():
+    r"""Ensure progress bar is refreshed.
+
+    If the progress bar has to wait for a long time
+    until it would get updated,
+    we enforce an update by a given time.
+
+    """
+    for _ in audeer.progress_bar(range(2), maximum_refresh_time=0.01):
+        time.sleep(0.05)
--- a/audeer/core/tqdm.py
+++ b/audeer/core/tqdm.py
@@ -49,7 +49,7 @@ def progress_bar(
     total: int = None,
     desc: str = None,
     disable: bool = False,
-    maximum_refresh_time: typing.Optional = 1,
+    maximum_refresh_time: float = None,
 ) -> tqdm:
     r"""Progress bar with optional text on the right.
 
@@ -107,10 +107,32 @@ def progress_bar(
     )
 
 
-def tqdm_wrapper(iterable, maximum_refresh_time, *args, **kwargs):
-    r"""Progress bar wrapper to enforce update once a second.
+def tqdm_wrapper(
+    iterable: typing.Sequence,
+    maximum_refresh_time: float,
+    *args,
+    **kwargs,
+) -> tqdm:
+    r"""Tqdm progress bar wrapper to enforce update once a second.
 
-    See https://github.com/tqdm/tqdm/issues/861#issuecomment-2197893883.
+    When using tqdm with large time durations
+    between single steps of the iteration,
+    it will not automatically update the elapsed time,
+    but needs to be forced,
+    see https://github.com/tqdm/tqdm/issues/861#issuecomment-2197893883.
+
+    Args:
+        iterable: sequence to iterate through
+        maximum_refresh_time: refresh the progress bar
+            at least every ``maximum_refresh_time`` seconds,
+            using another thread.
+            If ``None``,
+            no refreshing is enforced
+        args: arguments passed on to ``tqdm``
+        kwargs: keyword arguments passed on to ``tqdm``
+
+    Returns:
+        progress bar object
 
     """
     pbar = tqdm(iterable, *args, **kwargs)
--- a/audeer/core/tqdm.py
+++ b/audeer/core/tqdm.py
@@ -1,4 +1,6 @@
-from typing import Sequence
+import threading
+import time
+import typing
 
 from tqdm import tqdm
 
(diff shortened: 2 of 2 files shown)


tests/test_tqdm.py
```diff
@@ -37,3 +37,15 @@ def test_progress_bar():
...
+def test_progress_bar_update():
+    r"""Ensure progress bar is refreshed.
+
+    If the progress bar has to wait for a long time
+    until it would get updated,
+    we enforce an update by a given time.
+
+    """
+    for _ in audeer.progress_bar(range(2), maximum_refresh_time=0.01):
+        time.sleep(0.05)
```
hagenw, line 51: I do not test the display output of the progress bar here, but only ensure the code is executed in the tests.

We decided against testing the expected output of the progress bar when we introduced it, as it was not straightforward to achieve and `tqdm` is a well established library. What we do instead is manual tests of the expected behavior.

ChristianGeng: Understood. you keep `maximum_refresh_time`at a small value to override the (longer) sleep interval such you can be sure that the refresh thread callback in `def refresh` is covered by the test. 

This is ok with me. I am resolving this.

ChristianGeng (commented):
# Review

The MR allows to add an additional param that will refresh the progress bar even when an iteration takes confusingly wrong. Like this the user can be more confident, seeing that the pbar is still alive. 

## Threading

I will be upfront with slappping the label "noob" onto my knowledge level  when it comes to threading .-).  Saying that I *think* I understand that the `refresh` thread specified as a daemon thread does not have to be joined, as it will die anyway when the main thread exits, and does not continue existing as a zombie.
At least [this comment](https://stackoverflow.com/questions/4330111/meaning-of-daemon-property-on-python-threads#:~:text=%40Falmarri%3A%20What%20has,Dec%202%2C%202010) to me suggtests that this is ok.

## Implementation 

I wonder how this relates to the parameters args of `tqdm.tqdm` , see [here](https://tqdm.github.io/docs/tqdm/#:~:text=mininterval%3A%20float%2C%20optional,set%20miniters%3D1.)? 
Would not the parameters `miniters, mininterval, maxinterval`  allow to create identical or at least similar behavior?

hagenw:
> I wonder how this relates to the parameters args of tqdm.tqdm , see [here](https://tqdm.github.io/docs/tqdm/#:~:text=mininterval%3A%20float%2C%20optional,set%20miniters%3D1.)?
> Would not the parameters miniters, mininterval, maxinterval allow to create identical or at least similar behavior?

Yes, you would expect it, but I never got `maxinterval` to work. There are also issues about that, see https://github.com/tqdm/tqdm/issues/861#issuecomment-562825165 and https://github.com/tqdm/tqdm/issues/717. The implementation I used here was presented as a possible solution to the first issue at https://github.com/tqdm/tqdm/issues/861#issuecomment-2197893883

hagenw:
The better solution would be to fix `tqdm` to respect `maxinterval`, but I'm too lazy to figure out how to do it ;)

ChristianGeng:
>> The better solution would be to fix tqdm to respect maxinterval, but I'm too lazy to figure out how to do it ;)

With the problems that exist using m̀axinterval` I would say that the implementation is ok.

> Honestly, I also have no clue. I will try to investigate this first, before merging here.

In anticipation I will approve the MR in order to apply my new knowledge on how to re-revie and then approve ,-)

ChristianGeng (approved):
Redoing the review, this MR is approved as 

- the discussion about the `tqdm` has been settled with the result that this is a workaround needed to overcome the current tqdm status
- the `daemon` parameter of the refresh callback is being checked and can be fixed later in the unlikely case this is needed.

hagenw:
I tested if the thread goes away with the following code
```python
import threading
import time

import audeer


print(f"Number of threads before progress bar: {threading.active_count()}")
print("Active threads:")
for thread in threading.enumerate():
    print(thread)
print()
for _ in audeer.progress_bar(range(2), maximum_refresh_time=0.01):
    time.sleep(0.05)
    print(f"Number of threads during progress bar: {threading.active_count()}")
    print("Active threads:")
    for thread in threading.enumerate():
        print(thread)
print()
print(f"Number of threads after progress bar: {threading.active_count()}")
print("Active threads:")
for thread in threading.enumerate():
    print(thread)
```
We get one additional thread, when starting with a fresh Python console, after the progress bar, but this is a generic monitoring thread. The thread that refreshes the progress bar is only active when the progress bar gets updated:
```
Number of threads before progress bar: 1
Active threads:
<_MainThread(MainThread, started 135948201152512)>

Number of threads during progress bar: 3
<_MainThread(MainThread, started 135948201152512)>
<TMonitor(Thread-1, started daemon 135948176623168)>
<Thread(Thread-2 (refresh), started daemon 135948168230464)>
Number of threads during progress bar: 3
<_MainThread(MainThread, started 135948201152512)>
<TMonitor(Thread-1, started daemon 135948176623168)>
<Thread(Thread-2 (refresh), started daemon 135948168230464)>

Number of threads after progress bar: 2
Active threads:
<_MainThread(MainThread, started 135948201152512)>
<TMonitor(Thread-1, started daemon 135948176623168)>
```

PR #150, opened 2024-07-05
DOC: fix return value of run_tasks()

hagenw:
This fixes the docstring of `audeer.run_tasks()` by changing its return type from sequence to list, and adds missing documentation of its return value.

image


ChristianGeng (approved):
This is very similar to #149 : This is definitively an improvement as it is more specific about the return type that is always list. 

Approving.

PR #152, opened 2024-07-05
Release 2.1.0

hagenw:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,22 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 2.1.0 (2024-07-05)
+--------------------------
+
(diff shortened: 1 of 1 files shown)


ChristianGeng (approved):
I believe that I have reviewed all changes in this release's component releases. 

I have had a look through the diff to the last tag, and for audeer this is probably not relevant but I see that the build workflow  now does this:

```
python-version: [ '3.10' ]
```
As said, here I will approve right away. But in other packages, should the release log not  mention when a specific python version is deprecated? I remember that I hadn't complained about this in some `audformat/audb` releases recently. But do you think  this should be mentioend in CHANGELOGS in general?

As said this is a general question, and approval is given.

hagenw:
We still test for all Python versions:

image

For the build workflow we just have to pick one Python version to create the environment that builds the Python package.

PR #153, opened 2024-07-12
Use smoothing=0 in audeer.progress_bar()

hagenw:
Closes https://github.com/audeering/audb/issues/442

This makes the estimation of the remaining time in the progress bar less eratic.

For a single thread it changed the behavior from

bar-main

to

bar

And when using 4 threads it changes the behavior from

bar-main-num-workers

to

bar-num-workers

### DIFF
--- a/audeer/core/tqdm.py
+++ b/audeer/core/tqdm.py
@@ -104,6 +104,7 @@ def progress_bar(
         disable=disable,
         desc=format_display_message(desc, pbar=True),
         leave=config.TQDM_LEAVE,
+        smoothing=0,
     )
 
 


ChristianGeng (approved):
This is an exponential smoothing parameter.  Setting it to 0 afaics seems to result in a mean estimate of all previous iteration durations.  


This is safe and can be approved.

hagenw:
> This is safe and can be approved.

It's always complicated with judging all the effects a setting on the progress bar has. Sometimes you realize only after a few days of using it, that it has some unwanted consequences. But the good news is, it will not result in any failing code ;)

If we really find a strange side effect of this, we can revert the change in a later release.

PR #155, opened 2024-07-18
Add audeer.script_dir()

hagenw:
Very often, I add the following to the top of a script:

```python
CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
```
and I need to google it every time.

To make it easier to remember, this pull request proposes to add `audeer.script_dir()` which can be used instead.
When called in an interactive session, it returns the directory the user is located in.

image

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -1,6 +1,7 @@
 import errno
 import fnmatch
 import hashlib
+import inspect
 import itertools
 import os
 import platform
@@ -229,6 +230,21 @@ def create_archive(
         raise RuntimeError(
             f"You can only create a ZIP or TAR.GZ archive, " f"not {archive}"
         )
+
+
+def current_dir() -> str:
+    r"""Folder in which caller of this function is located.
+
+    Returns:
+        current directory of caller
+
+    Examples:
+        >>> os.path.basename(current_dir())
+        'audeer_core_io_current_dir0'
+
+    """
+    caller = inspect.stack()[1].filename
+    return os.path.dirname(os.path.realpath(caller))
 
 
 def download_url(
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -232,30 +232,6 @@ def create_archive(
         )
 
 
-def current_dir() -> str:
-    r"""Folder in which caller of this function is located.
-
-    When called from a file,
-    it returns the directory,
-    in which the file is stored.
-    When called in an interactive session,
-    it returns the current directory
-    of the interactive session.
-
-    Returns:
-        current directory of caller
-
-    Examples:
-        >>> os.path.basename(current_dir())  # folder of docstring test
-        'audeer_core_io_current_dir0'
-
-    """
-    # See https://stackoverflow.com/a/37792573
-    caller = inspect.stack()[1].filename
-    # See https://stackoverflow.com/a/5137509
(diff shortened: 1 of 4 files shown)


frankenjoe:
Haha, just this morning I had the very same idea that such a function would be a good extension to `audeerr` :)

hagenw:
Hehe, great, I still wonder why I hadn't the idea eralier.

audeer/core/io.py
```diff
@@ -231,6 +232,30 @@ def create_archive(
...
 
 
+def current_dir() -> str:
+    r"""Folder in which caller of this function is located.
+
+    When called from a file,
+    it returns the directory,
+    in which the file is stored.
+    When called in an interactive session,
+    it returns the current directory
```
hagenw, line 242: Usually, this is called "current working directory", e.g. https://docs.python.org/3/library/os.html#os.getcwd.
Should I change this here as well?

frankenjoe:
I wonder if the name `current_dir()` could be misleading. A user might expect it returns the current working directory. So maybe we should rename to `script_dir()`?

hagenw:
I found `current_dir()` a nice fit, as the variable name in the script was always `CURRENT_DIR`. But you are right, inside a script it might be misleading as it returns the script dir. On the other hand, in an interactive session it returns the current working directory.
So, it is actually a mixture of `script_dir()` and `current_working_dir()`. A correct term might be `caller_dir()`, but I doubt that this name is a good choice.

frankenjoe:
>  On the other hand, in an interactive session it returns the current working directory.

You have already documented the behavior when no script file exists, e.g. in an interactive session. So I think it is ok to still call it `script_dir()` or maybe `current_script_dir()` to align it with `current_working_dir()`.

hagenw:
I changed the name to `audeer.script_dir()` and extended the tests to check that a change in current working directory, does not affect `audeer.script_dir()` when called in a file.

audeer/core/io.py
frankenjoe: Yes, I would name it "current working directory".

hagenw: I changed that already, but forgot to update the screen shot accordingly. I have fixed that now.

audeer/core/io.py
```diff
@@ -1007,6 +1008,30 @@ def rmdir(
...
+    Returns:
+        current directory of caller
+
+    Examples:
+        >>> os.path.basename(script_dir())  # folder of docstring test
+        'audeer_core_io_script_dir0'
+
+    """
+    # See https://stackoverflow.com/a/37792573
+    caller = inspect.stack()[1].filename
```
frankenjoe, line 1030: I didn't find the link on stackoverflow very helpful to understand why we cannot use `__file__` here. Maybe you can in addition add an explaining sentence here?

hagenw: Yes, good point.
I changed to
```python
    # Returning the script dir is usually done with
    # `os.path.dirname(os.path.realpath(__file__))`,
    # see https://stackoverflow.com/a/5137509.
    # We cannot use `__file__` here,
    # as this would always point to this file (`io.py`).
    # Instead we find the script
    # of the caller of `audeer.script_dir()`,
    # see https://stackoverflow.com/a/37792573
    caller = inspect.stack()[1].filename
    return os.path.dirname(os.path.realpath(caller))
```

PR #157, opened 2024-07-23
Release 2.2.0

hagenw:
image

For reference, `audeer.script_dir()` was added in https://github.com/audeering/audeer/pull/155

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,8 +11,8 @@ Version 2.2.0 (2024-07-23)
 --------------------------
 
 * Added: ``audeer.script_dir()``
-  which returns the directory of the file,
-  in which the function is called,
+  which returns the directory
+  of the file where it was called,
   or the current working directory
   when called in an interactive session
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,16 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 2.2.0 (2024-07-23)
+--------------------------
+
+* Added: ``audeer.script_dir()``
+  which returns the directory of the file,
+  in which the function is called,
```
ChristianGeng, line 15: I would say "  returns the directory of the file where it is called" or " returns the directory of the file from where it is called". I mean the "in which" that sounds strange to me.

CHANGELOG.rst
```diff
@@ -7,6 +7,16 @@ The format is based on `Keep a Changelog`_,
...
+  or the current working directory
+  when called in an interactive session
```
ChristianGeng, line 17: Having no full stop at the end of a sentence is more common when it comes to CHANGELOG entries.

ChristianGeng (approved):
There are only cosmetic changes necessary. 

Approving straight away.

CHANGELOG.rst
hagenw: Good idea, I was also not happy with having two times "which". I changed to your first suggestion.

PR #166, opened 2025-01-07
Support all tar archives in create and extract

hagenw:
This adds support to not only create and extract `tar.gz` files, but all other supported `tar` files as well.

In addition, `audeer.create_archive()` and `audeer.extract_archive()` were slightly refactored to improve the code quality.

Updated docstring of `audeer.create_archive()`:

image

Updated docstring of `audeer.extract_archive()`:

image

Updated docstring of `audeer.extract_archives()`:

image


tests/test_io.py
```diff
@@ -116,6 +116,15 @@ def tree(tmpdir, request):
             ".",
```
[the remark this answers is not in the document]
hagenw: We already have a test for this case.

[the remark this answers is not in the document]
hagenw: But maybe it is a good idea to add also some common used archive extensions like `7z` and `rar`. I will update the tests.

audeer/core/io.py
```diff
@@ -351,39 +362,35 @@ def extract_archive(
...
+            members = zf.infolist()
+            for member in progress_bar(members, desc=desc, disable=disable):
+                zf.extract(member, destination)
+            return [m.filename for m in members]
+
+    def extract_tar(archive: str) -> list:
+        with tarfile.open(archive, "r") as tf:
+            members = tf.getmembers()
+            for member in progress_bar(members, desc=desc, disable=disable):
+                # In Python 3.12 the `filter` argument was introduced,
```
ChristianGeng, line 376: The value of the filter arg we have discussed previously, and it needn't be reiterated again in this MR.  Please correct me if I'm wrong here.

ChristianGeng: Then this can be resolved and it needn't be resumed here.

audeer/core/io.py
```diff
@@ -211,21 +212,31 @@ def create_archive(
...
-            for file in progress_bar(files, desc=desc, disable=disable):
-                full_file = safe_path(root, file)
-                zf.write(full_file, arcname=file)
-    elif archive.endswith("tar.gz"):
-        with tarfile.open(archive, "w:gz") as tf:
-            for file in progress_bar(files, desc=desc, disable=disable):
-                full_file = safe_path(root, file)
-                tf.add(full_file, file)
-    else:
+    archive_handlers = {
```
ChristianGeng, line 215: My understanding is that these archive handlers are wisely chosen:
After doing some research I found

- that "tgz" and "tbz" are only there for historical reasons - with windows only wanting a single extension of length 3. 

- 7-zip has high compression rate it is said to be slow, has external dependencies that might not be installed. 

So I think that it is fully ok to not provide these by default.
 
What I wonder though: in case an impatient user needs to patch an archive handler for a new compression that `audeer` cannot integrate fast enough: would it make sense to move these handlers into something like a  `define`  module such that the handler can be patched in an ad-hoc fashion by the impatient?

hagenw: Would be great if it would work that way, but I'm afraid that it is not enough to define how to open the file as even the existing archive handlers have already a different syntax to add a file to the archive:

```python
            if extension == "zip":
                archive_file.write(full_file, arcname=file)
            else:
                archive_file.add(full_file, file)
```

So I guess to make it easier extendable, one would need to introduce something like a base classe.

I would propose to stay here with the current solution, but maybe open an issue requesting to make it easier to integrate custom or other archive handlers.

ChristianGeng: Yes, sure, `gzip` and `bzip` do not handle archiving this is why they resort to `tar`  for doing the achieving, this is why it is being called  `add` as opposed to `write` in zip.

I would not oppose having a base class or equivalent mechanism to handle adding/writing.
But the argument that this is beyond scope here is enough to suggest that we can  resolve here.

ChristianGeng (approved):
Single issues have been discussed in separate threads. 
As these are resolved I think that it will be safe to proceed and give approval.

PR #174, opened 2025-10-02
Add audeer.suppress_stdout()

hagenw:
Add `audeer.suppress_stdout()`. It's useful to suppress output of code that generates a lot of meaningless output which is not configurable by the user, e.g. external machine learning models.

image

## Summary by Sourcery

Introduce a context manager to temporarily suppress standard output during code execution.

New Features:
- Add audeer.suppress_stdout() context manager to silence stdout

Tests:
- Add tests to verify suppression of output, nested contexts, exception handling, restoration of stdout, and preservation of stderr

### DIFF
--- a/audeer/__init__.py
+++ b/audeer/__init__.py
@@ -33,6 +33,7 @@ from audeer.core.utils import is_uid
 from audeer.core.utils import run_tasks
 from audeer.core.utils import run_worker_threads
 from audeer.core.utils import sort_versions
+from audeer.core.utils import suppress_stdout
 from audeer.core.utils import to_list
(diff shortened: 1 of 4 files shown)


frankenjoe:
I tried:

```python
import audeer
import audb


with audeer.suppress_stdout():
    audb.load("emodb")
```

And it indeed did not show:

```
Get:   emodb v2.0.0
Cache: /home/jwagner/data/audb/emodb/2.0.0/d3b62a9b
```

but only the progress bar, which makes sense, since this is printed to `stderr` I think.

So looks all good to me!

PR #184, opened 2025-11-11
Add num_workers argument to extract_archive()

hagenw:
This speeds up extracting of archives containing multiple files by adding a `num_workers` argument.

image

## Summary by Sourcery

Enable configurable parallel extraction in extract_archive by introducing a num_workers parameter, refactoring member extraction, and adding tests to ensure multi-threaded behavior matches sequential results.

New Features:
- Add num_workers parameter to extract_archive to enable parallel extraction of archive members
- Support parallel extraction of ZIP and TAR archives using multiple workers

Enhancements:
- Refactor extraction logic into per-member helper functions and unify sequential and parallel flows
- Allow num_workers=None to default to CPU count multiplied by 5 for optimal parallelism

Documentation:
- Update extract_archive docstring to describe the new num_workers argument

Tests:
- Introduce parameterized tests verifying multi-threaded extraction yields identical results to single-threaded extraction for ZIP and TAR archives

PR #186, opened 2025-11-19
Multithreading archive extraction

hagenw:
Alternative implementation to https://github.com/audeering/audeer/pull/184.

This adds the `num_workers` argument to `audeer.extract_archive()` and `audeer.extract_archives()`.

**Benchmark**

I run a benchmark on a 1GB ZIP file containing 1000 files, the results are

| num_workers | Time (s) |
| --- | --- |
| 1 | 1.5 |
| 10 | 2.0 |

We do not really get any speedup by using multithreading for extracting the archive.

## Summary by Sourcery

Enable parallel extraction of ZIP and TAR archives by adding a num_workers parameter to extract_archive and extract_archives.

New Features:
- Add num_workers argument to extract_archive to split and extract archive members in parallel.
- Add num_workers argument to extract_archives to concurrently extract multiple archives.

Enhancements:
- Refactor extract_zip and extract_tar to accept optional member lists for chunked extraction.
- Introduce extract_chunk helper and integrate run_tasks for concurrent extraction workflows.

Tests:
- Add tests for multithreaded extraction of ZIP and TAR archives and for extracting multiple archives with num_workers > 1.

### DIFF
--- a/audeer/core/io.py
+++ b/audeer/core/io.py
@@ -15,7 +15,6 @@ import zipfile
 from audeer.core.path import path as safe_path
 from audeer.core.tqdm import format_display_message
 from audeer.core.tqdm import progress_bar
-from audeer.core.utils import flatten_list
 from audeer.core.utils import run_tasks
 from audeer.core.utils import to_list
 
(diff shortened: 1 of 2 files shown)


hagenw:
Adding multi-threading to archive extraction does not provide any speedups.

Instead we should consider extracting archives in parallel (https://github.com/audeering/audeer/issues/187) or when downloading the archive from a backend use streaming ZIP extraction to extract the archive while downloading (https://github.com/audeering/audbackend/pull/279).

PR #196, opened 2026-07-21
Deep merge, explicit types in load_configuration()

ChristianGeng:
Extends `load_configuration()` (#195) with three features needed to assure broad applicability
of `load_settings()` in other projects, where config is organised in
nested sections. 

Three commits:

1. **Deep-merge nested sections** — a user file overrides only the keys it
   defines instead of replacing the whole section (lists still replace whole).
2. **Nested env overrides** — levels joined by `__`, e.g.
   `PKG_MODEL__DEVICE` → `model.device` (pydantic-settings convention).
3. **Explicit `types=`** — declares the target type for `None`-defaulted keys
   so env overrides cast correctly; no new dependency.

Stacked on `config-handling-helper`; will retarget to `main` after #195 lands.
Purely additive — existing tests unchanged, full suite green, 100% coverage.

### DIFF
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -43,6 +43,16 @@ def load_configuration(
     by joining the levels with ``__``,
     e.g. ``model.device``
     is overridden by ``<env_prefix>_MODEL__DEVICE``.
+    A whole nested mapping can instead be replaced
+    by a single variable holding a JSON object,
+    e.g. ``<env_prefix>_MODEL='{"device": "cpu"}'``;
+    the object replaces the mapping as a whole
+    (keys it omits are dropped)
+    and may introduce keys not present in the files.
+    Nested variables are applied afterwards
+    and therefore take precedence,
+    e.g. ``<env_prefix>_MODEL__DEVICE``
+    overrides ``device`` from ``<env_prefix>_MODEL``.
     The value of an environment variable is converted
     to the type of the corresponding default value:
     ``str`` values are used as they are,
@@ -248,7 +258,17 @@ def _override_with_environment(
                     f"The 'types' entry for the nested section '{key}' "
                     f"must be a mapping, but is '{type(key_type).__name__}'."
                 )
-            _override_with_environment(default_value, name, key_type or {}, "__")
+            # A whole-section variable (e.g. PKG_MODEL) replaces the mapping
+            # as a JSON object; nested variables (e.g. PKG_MODEL__DEVICE) are
+            # applied afterwards and therefore take precedence.
+            if name in os.environ:
+                cfg[key] = _parse_environment_value(
+                    name,
+                    os.environ[name],
+                    default_value,
+                    dict,
+                )
+            _override_with_environment(cfg[key], name, key_type or {}, "__")
         elif name in os.environ:
             cfg[key] = _parse_environment_value(
                 name,
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -67,6 +67,9 @@ def load_configuration(
     raise a ``ValueError`` on invalid input.
     Only keys already present in the configuration files
     can be overridden by environment variables.
+    Only string keys are matched;
+    a non-string key (e.g. a numeric YAML key)
+    is left untouched.
 
     A default value of ``None`` carries no type,
     so an environment variable would be kept as a string.
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -118,9 +118,6 @@ def load_configuration(
 
         A key that defaults to ``None`` has no inferred type,
         so declare it in ``types``.
-        The environment variable then holds
-        the JSON representation of the value,
-        e.g. a JSON array for a ``list``:
 
         >>> import os
         >>> config_file = audeer.path(tempfile.mkdtemp(), "config.yaml")
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -250,9 +250,11 @@ def _validate_types(cfg: Mapping, types: Mapping) -> None:
 
     """
     for key, value in cfg.items():
-        declared = types.get(key)
-        if declared is None:
+        # An explicit ``None`` entry is a malformed declaration,
+        # only an absent key means "no type declared"
+        if key not in types:
             continue
+        declared = types[key]
         if isinstance(value, Mapping):
             if not isinstance(declared, Mapping):
                 raise ValueError(
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -38,6 +38,10 @@ def load_configuration(
     prefixed with ``env_prefix`` and an underscore,
     e.g. the key ``cache_root``
     is overridden by ``<env_prefix>_CACHE_ROOT``.
+    Keys inside nested mappings are addressed
+    by joining the levels with ``__``,
+    e.g. ``model.device``
+    is overridden by ``<env_prefix>_MODEL__DEVICE``.
     The value of an environment variable is converted
     to the type of the corresponding default value:
     ``str`` values are used as they are,
@@ -178,11 +182,28 @@ def _load_configuration_file(config_file: str) -> dict:
 def _override_with_environment(
     cfg: dict,
     env_prefix: str,
+    separator: str = "_",
 ) -> None:
-    r"""Override configuration values with environment variables in place."""
+    r"""Override configuration values with environment variables in place.
+
+    Nested mappings are traversed recursively.
+    The environment variable name of a nested key
+    joins the levels with ``__``,
+    e.g. ``model.device`` with prefix ``PKG``
+    is overridden by ``PKG_MODEL__DEVICE``.
+
+    Args:
+        cfg: configuration dictionary, modified in place
+        env_prefix: name prefix accumulated so far
+        separator: string joining ``env_prefix`` and the current key
+            (``"_"`` at the top level, ``"__"`` for nested levels)
+
+    """
     for key, default_value in cfg.items():
-        name = f"{env_prefix}_{key.upper()}"
-        if name in os.environ:
+        name = f"{env_prefix}{separator}{key.upper()}"
+        if isinstance(default_value, Mapping):
+            _override_with_environment(default_value, name, "__")
+        elif name in os.environ:
             cfg[key] = _parse_environment_value(
                 name,
                 os.environ[name],
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -89,7 +89,7 @@ def load_configuration(
     ``types`` mirrors the (possibly nested) structure
     of the configuration;
     an entry that does not match a configuration key
-    raises a ``ValueError`` to catch misspellings.
+    raises a ``ValueError``.
 
     Missing or empty configuration files are skipped.
 
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -250,6 +250,10 @@ def _override_with_environment(
 
     """
     for key, default_value in cfg.items():
+        # Environment variables can only address string keys;
+        # non-string keys (e.g. integers) are left untouched.
+        if not isinstance(key, str):
+            continue
         name = f"{env_prefix}{separator}{key.upper()}"
         key_type = types.get(key)
         if isinstance(default_value, Mapping):
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -116,6 +116,23 @@ def load_configuration(
         >>> audeer.load_configuration(config_file)
         {'cache_root': '~/cache'}
 
+        A key that defaults to ``None`` has no inferred type,
+        so declare it in ``types``.
+        The environment variable then holds
+        the JSON representation of the value,
+        e.g. a JSON array for a ``list``:
+
+        >>> import os
+        >>> config_file = audeer.path(tempfile.mkdtemp(), "config.yaml")
+        >>> with open(config_file, "w") as file:
+        ...     _ = file.write("hosts: null\n")
+        >>> os.environ["APP_HOSTS"] = '["host1", "host2"]'
+        >>> audeer.load_configuration(
+        ...     config_file, env_prefix="APP", types={"hosts": list}
+        ... )
+        {'hosts': ['host1', 'host2']}
+        >>> del os.environ["APP_HOSTS"]
+
     """
     cfg = _load_configuration_file(default_config_file)
 
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -149,8 +149,12 @@ def load_configuration(
         for user_config_file in user_config_files:
             _deep_merge(cfg, _load_configuration_file(user_config_file))
 
-    if types is not None and not isinstance(types, Mapping):
-        raise ValueError(f"'types' must be a mapping, but is '{type(types).__name__}'.")
+    if types is not None:
+        if not isinstance(types, Mapping):
+            raise ValueError(
+                f"'types' must be a mapping, but is '{type(types).__name__}'."
+            )
+        _validate_types(cfg, types)
 
     if env_prefix is not None:
         _override_with_environment(cfg, env_prefix, types=types or {})
@@ -226,6 +230,39 @@ def _load_configuration_file(config_file: str) -> dict:
     return dict(cfg)
 
 
+def _validate_types(cfg: Mapping, types: Mapping) -> None:
+    r"""Validate declared ``types`` against the configuration structure.
+
+    Checks, independent of which environment variables are set, that a declared
+    leaf type is a class and a declared section type is a mapping. Only entries
+    that mirror the configuration are considered.
+
+    Args:
+        cfg: configuration dictionary
+        types: declared types mirroring ``cfg``
+
+    Raises:
+        ValueError: if a declared section type is not a mapping
+        ValueError: if a declared leaf type is not a class
+
+    """
+    for key, value in cfg.items():
+        declared = types.get(key)
+        if declared is None:
+            continue
+        if isinstance(value, Mapping):
+            if not isinstance(declared, Mapping):
+                raise ValueError(
+                    f"The 'types' entry for the nested section '{key}' "
+                    f"must be a mapping, but is '{type(declared).__name__}'."
+                )
+            _validate_types(value, declared)
+        elif not isinstance(declared, type):
+            raise ValueError(
+                f"The 'types' entry for '{key}' is not a type: {declared!r}."
+            )
+
+
 def _override_with_environment(
     cfg: dict,
     env_prefix: str,
@@ -257,11 +294,6 @@ def _override_with_environment(
         name = f"{env_prefix}{separator}{key.upper()}"
         key_type = types.get(key)
         if isinstance(default_value, Mapping):
-            if key_type is not None and not isinstance(key_type, Mapping):
-                raise ValueError(
-                    f"The 'types' entry for the nested section '{key}' "
-                    f"must be a mapping, but is '{type(key_type).__name__}'."
-                )
             # A whole-section variable (e.g. PKG_MODEL) replaces the mapping
             # as a JSON object; nested variables (e.g. PKG_MODEL__DEVICE) are
             # applied afterwards and therefore take precedence.
@@ -304,11 +336,6 @@ def _parse_environment_value(
     """
     if target_type is None:
         target_type = type(default_value)
-    if not isinstance(target_type, type):
-        raise ValueError(
-            f"The type declared for the value overridden by '{name}' "
-            f"is not a type: {target_type!r}."
-        )
     try:
         # ``bool`` has to be checked before ``int``,
         # as ``bool`` is a subclass of ``int``.
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -98,7 +98,9 @@ def load_configuration(
             mirroring the (possibly nested) configuration structure.
             Used to cast environment variable overrides
             for keys whose default value is ``None``,
-            or to override the type inferred from the default value
+            or to override the type inferred from the default value.
+            Supported types are
+            ``bool``, ``int``, ``float``, ``str``, ``list``, ``dict``
         validate: callable that receives the merged configuration
             dictionary and raises an error if it is invalid.
             It is applied once,
@@ -116,7 +118,8 @@ def load_configuration(
             cannot be converted to the type
             of the corresponding default value
         ValueError: if a type declared in ``types``
-            is not a class
+            is not a class,
+            or not one of the supported types
         ValueError: if ``types``,
             or a ``types`` entry for a nested section,
             is not a mapping
@@ -265,6 +268,15 @@ def _validate_types(cfg: Mapping, types: Mapping) -> None:
         elif not isinstance(declared, type):
             raise ValueError(
                 f"The 'types' entry for '{key}' is not a type: {declared!r}."
+            )
+        elif not issubclass(declared, (bool, int, float, str, list, dict)):
+            # Anything else would silently fall through
+            # to keeping the environment variable a string
+            raise ValueError(
+                f"The 'types' entry for '{key}' "
+                f"is not a supported type: '{declared.__name__}'. "
+                f"Supported types are "
+                f"'bool', 'int', 'float', 'str', 'list', 'dict'."
             )
 
 
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -10,6 +10,7 @@ def load_configuration(
     user_config_files: str | Sequence[str] | None = None,
     *,
     env_prefix: str | None = None,
+    types: Mapping | None = None,
     validate: Callable[[dict], None] | None = None,
 ) -> dict:
     r"""Load configuration from files and environment variables.
@@ -57,6 +58,14 @@ def load_configuration(
     Only keys already present in the configuration files
     can be overridden by environment variables.
 
+    A default value of ``None`` carries no type,
+    so an environment variable would be kept as a string.
+    Use ``types`` to declare the intended type
+    for such keys,
+    or to override the inferred type of any key.
+    ``types`` mirrors the (possibly nested) structure
+    of the configuration.
+
     Missing or empty configuration files are skipped.
 
     Reading configuration files requires ``pyyaml``,
@@ -71,6 +80,12 @@ def load_configuration(
         env_prefix: prefix of environment variables
             used to override configuration values.
             If ``None``, environment variables are ignored
+        types: mapping that declares the type
+            of configuration values,
+            mirroring the (possibly nested) configuration structure.
+            Used to cast environment variable overrides
+            for keys whose default value is ``None``,
+            or to override the type inferred from the default value
         validate: callable that receives the merged configuration
             dictionary and raises an error if it is invalid.
             It is applied once,
@@ -106,7 +121,7 @@ def load_configuration(
             _deep_merge(cfg, _load_configuration_file(user_config_file))
 
     if env_prefix is not None:
-        _override_with_environment(cfg, env_prefix)
+        _override_with_environment(cfg, env_prefix, types=types or {})
 
     if validate is not None:
         validate(cfg)
@@ -182,6 +197,7 @@ def _load_configuration_file(config_file: str) -> dict:
 def _override_with_environment(
     cfg: dict,
     env_prefix: str,
+    types: Mapping,
     separator: str = "_",
 ) -> None:
     r"""Override configuration values with environment variables in place.
@@ -195,19 +211,24 @@ def _override_with_environment(
     Args:
         cfg: configuration dictionary, modified in place
         env_prefix: name prefix accumulated so far
+        types: declared types mirroring ``cfg``,
+            used to cast values whose default is ``None``
         separator: string joining ``env_prefix`` and the current key
             (``"_"`` at the top level, ``"__"`` for nested levels)
 
     """
     for key, default_value in cfg.items():
         name = f"{env_prefix}{separator}{key.upper()}"
+        key_type = types.get(key)
         if isinstance(default_value, Mapping):
-            _override_with_environment(default_value, name, "__")
+            nested_types = key_type if isinstance(key_type, Mapping) else {}
+            _override_with_environment(default_value, name, nested_types, "__")
         elif name in os.environ:
             cfg[key] = _parse_environment_value(
                 name,
                 os.environ[name],
                 default_value,
+                key_type,
             )
 
 
@@ -215,26 +236,36 @@ def _parse_environment_value(
     name: str,
     value: str,
     default_value: object,
+    target_type: type | None = None,
 ) -> object:
-    r"""Convert an environment variable to the type of the default value."""
+    r"""Convert an environment variable to the wanted type.
+
+    The target type is ``target_type`` if given,
+    otherwise the type of ``default_value``.
+    A ``None`` default without a declared type
+    leaves the value unchanged as a string.
+    """
+    if target_type is None:
+        target_type = type(default_value)
     try:
-        # ``bool`` has to be checked before ``int``.
+        # ``bool`` has to be checked before ``int``,
+        # as ``bool`` is a subclass of ``int``.
         # A boolean never fails conversion:
         # any value other than the truthy ones becomes ``False``
-        if isinstance(default_value, bool):
+        if issubclass(target_type, bool):
             return value.lower() in ("1", "true", "yes", "on")
-        if isinstance(default_value, int):
+        if issubclass(target_type, int):
             return int(value)
-        if isinstance(default_value, float):
+        if issubclass(target_type, float):
             return float(value)
         # ``json.JSONDecodeError`` is a subclass of ``ValueError``
-        if isinstance(default_value, (list, dict)):
+        if issubclass(target_type, (list, dict)):
             return json.loads(value)
     except ValueError as ex:
         raise ValueError(
             f"The environment variable '{name}={value}' "
             f"could not be converted to the type "
-            f"'{type(default_value).__name__}' "
+            f"'{target_type.__name__}' "
             f"of the corresponding configuration value."
         ) from ex
     return value
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -443,6 +443,11 @@ def _override_with_environment(
                 default_value,
                 key_type,
             )
+            if isinstance(cfg[key], Mapping):
+                # A mapping introduced via a declared ``dict`` type
+                # behaves like a section, so nested variables
+                # are applied on top of it as well
+                _override_with_environment(cfg[key], f"{name}__", {})
 
 
 def _parse_environment_value(
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -23,6 +23,8 @@ def load_configuration(
     2. ``user_config_files``,
        applied in the given order
        (a later file overrides an earlier one)
+    3. environment variables,
+       if ``env_prefix`` is given
 
     Configuration files are deep-merged:
     nested mappings are merged key by key,
@@ -31,8 +33,6 @@ def load_configuration(
     and keeps the remaining keys from the default.
     Any non-mapping value (including a list)
     replaces the previous value as a whole.
-    3. environment variables,
-       if ``env_prefix`` is given
 
     Environment variables are matched
     against the upper-cased configuration keys,
@@ -104,6 +104,9 @@ def load_configuration(
             of the corresponding default value
         ValueError: if a type declared in ``types``
             is not a class
+        ValueError: if ``types``,
+            or a ``types`` entry for a nested section,
+            is not a mapping
 
     Examples:
         >>> import tempfile
@@ -121,6 +124,9 @@ def load_configuration(
             user_config_files = [user_config_files]
         for user_config_file in user_config_files:
             _deep_merge(cfg, _load_configuration_file(user_config_file))
+
+    if types is not None and not isinstance(types, Mapping):
+        raise ValueError(f"'types' must be a mapping, but is '{type(types).__name__}'.")
 
     if env_prefix is not None:
         _override_with_environment(cfg, env_prefix, types=types or {})
@@ -223,8 +229,12 @@ def _override_with_environment(
         name = f"{env_prefix}{separator}{key.upper()}"
         key_type = types.get(key)
         if isinstance(default_value, Mapping):
-            nested_types = key_type if isinstance(key_type, Mapping) else {}
-            _override_with_environment(default_value, name, nested_types, "__")
+            if key_type is not None and not isinstance(key_type, Mapping):
+                raise ValueError(
+                    f"The 'types' entry for the nested section '{key}' "
+                    f"must be a mapping, but is '{type(key_type).__name__}'."
+                )
+            _override_with_environment(default_value, name, key_type or {}, "__")
         elif name in os.environ:
             cfg[key] = _parse_environment_value(
                 name,
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -102,6 +102,8 @@ def load_configuration(
         ValueError: if an environment variable
             cannot be converted to the type
             of the corresponding default value
+        ValueError: if a type declared in ``types``
+            is not a class
 
     Examples:
         >>> import tempfile
@@ -247,6 +249,11 @@ def _parse_environment_value(
     """
     if target_type is None:
         target_type = type(default_value)
+    if not isinstance(target_type, type):
+        raise ValueError(
+            f"The type declared for the value overridden by '{name}' "
+            f"is not a type: {target_type!r}."
+        )
     try:
         # ``bool`` has to be checked before ``int``,
         # as ``bool`` is a subclass of ``int``.
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -287,6 +287,11 @@ def _validate_types(cfg: Mapping, types: Mapping) -> None:
                     f"must be a mapping, but is '{type(declared).__name__}'."
                 )
             _validate_types(value, declared)
+        elif isinstance(declared, Mapping):
+            raise ValueError(
+                f"The 'types' entry for '{key}' declares a nested section, "
+                f"but the corresponding configuration value is not a mapping."
+            )
         elif not isinstance(declared, type):
             raise ValueError(
                 f"The 'types' entry for '{key}' is not a type: {declared!r}."
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -265,14 +265,21 @@ def _override_with_environment(
             # A whole-section variable (e.g. PKG_MODEL) replaces the mapping
             # as a JSON object; nested variables (e.g. PKG_MODEL__DEVICE) are
             # applied afterwards and therefore take precedence.
+            nested_types = dict(key_type or {})
             if name in os.environ:
+                # The JSON replacement loses the default types, so remember
+                # the original leaf types and reuse them when casting the
+                # nested overrides below (an explicit ``types`` entry wins).
+                for nested_key, nested_default in default_value.items():
+                    if not isinstance(nested_default, Mapping):
+                        nested_types.setdefault(nested_key, type(nested_default))
                 cfg[key] = _parse_environment_value(
                     name,
                     os.environ[name],
                     default_value,
                     dict,
                 )
-            _override_with_environment(cfg[key], name, key_type or {}, "__")
+            _override_with_environment(cfg[key], name, nested_types, "__")
         elif name in os.environ:
             cfg[key] = _parse_environment_value(
                 name,
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -160,7 +160,7 @@ def load_configuration(
         _validate_types(cfg, types)
 
     if env_prefix is not None:
-        _override_with_environment(cfg, env_prefix, types=types or {})
+        _override_with_environment(cfg, f"{env_prefix}_", types or {})
 
     if validate is not None:
         validate(cfg)
@@ -270,7 +270,6 @@ def _override_with_environment(
     cfg: dict,
     env_prefix: str,
     types: Mapping,
-    separator: str = "_",
 ) -> None:
     r"""Override configuration values with environment variables in place.
 
@@ -282,11 +281,11 @@ def _override_with_environment(
 
     Args:
         cfg: configuration dictionary, modified in place
-        env_prefix: name prefix accumulated so far
+        env_prefix: name prefix accumulated so far,
+            including the trailing separator
+            (``PKG_`` at the top level, ``PKG_MODEL__`` below)
         types: declared types mirroring ``cfg``,
             used to cast values whose default is ``None``
-        separator: string joining ``env_prefix`` and the current key
-            (``"_"`` at the top level, ``"__"`` for nested levels)
 
     """
     for key, default_value in cfg.items():
@@ -294,7 +293,7 @@ def _override_with_environment(
         # non-string keys (e.g. integers) are left untouched.
         if not isinstance(key, str):
             continue
-        name = f"{env_prefix}{separator}{key.upper()}"
+        name = f"{env_prefix}{key.upper()}"
         key_type = types.get(key)
         if isinstance(default_value, Mapping):
             # A whole-section variable (e.g. PKG_MODEL) replaces the mapping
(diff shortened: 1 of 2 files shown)


audeer/core/config.py
```diff
@@ -145,42 +197,75 @@ def _load_configuration_file(config_file: str) -> dict:
...
 ) -> object:
-    r"""Convert an environment variable to the type of the default value."""
+    r"""Convert an environment variable to the wanted type.
+
+    The target type is ``target_type`` if given,
+    otherwise the type of ``default_value``.
+    A ``None`` default without a declared type
+    leaves the value unchanged as a string.
+    """
+    if target_type is None:
```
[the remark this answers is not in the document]
ChristianGeng: Fixed in 5434c4f. `_parse_environment_value` now guards `isinstance(target_type, type)` and raises a clear `ValueError` naming the env variable when a value declared in `types` is not a class, instead of letting `issubclass()` raise `TypeError`. Covered by `test_load_configuration_environment_types_not_a_type`.

tests/test_config.py
```diff
@@ -57,6 +57,23 @@ def test_load_configuration_user_file(tmpdir):
...
+def test_load_configuration_deep_merge(tmpdir):
+    default_file = write_config(
+        audeer.path(tmpdir, "default.yaml"),
+        "model:\n  device: cuda\n  lora: false\n",
+    )
+    user_file = write_config(
+        audeer.path(tmpdir, "user.yaml"),
+        "model:\n  lora: true\n",
+    )
+    # Nested sections are deep-merged:
```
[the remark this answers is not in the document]
ChristianGeng: Added `test_load_configuration_deep_merge_list_replaced` in 5434c4f: a list in the default is replaced wholesale by the user file's list, not merged element-wise.

tests/test_config.py
```diff
@@ -127,6 +144,23 @@ def test_load_configuration_environment(tmpdir, monkeypatch):
...
+    monkeypatch.setenv("PKG_MODEL__LORA", "true")
+    monkeypatch.setenv("PKG_GENERATION__TOP_K", "40")
+    config = audeer.load_configuration(config_file, env_prefix="PKG")
+    assert config == {
+        "model": {"device": "cpu", "lora": True},
+        "generation": {"top_k": 40},
+    }
+
+
 def test_load_configuration_environment_bool_false(tmpdir, monkeypatch):
```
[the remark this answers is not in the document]
ChristianGeng: Added `test_load_configuration_environment_nested_unknown_key` in 5434c4f: `PKG_MODEL__UNKNOWN` creates no entry, confirming only keys already present (including nested) can be overridden.

tests/test_config.py
```diff
@@ -137,6 +171,38 @@ def test_load_configuration_environment_bool_false(tmpdir, monkeypatch):
...
+def test_load_configuration_environment_types_none_default(tmpdir, monkeypatch):
+    config_file = write_config(
+        audeer.path(tmpdir, "default.yaml"),
+        "timeout: null\n",
+    )
+    monkeypatch.setenv("PKG_TIMEOUT", "2.5")
+    # A None default carries no type,
+    # so ``types`` declares the target type for the cast
+    config = audeer.load_configuration(
+        config_file,
```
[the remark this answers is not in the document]
ChristianGeng: Added `test_load_configuration_environment_types_override_default` in 5434c4f: with a string default, `types={'timeout': int}` casts the override to `int`, confirming `types` overrides the inferred type.

tests/test_config.py
```diff
@@ -137,6 +171,38 @@ def test_load_configuration_environment_bool_false(tmpdir, monkeypatch):
...
+def test_load_configuration_environment_types_nested(tmpdir, monkeypatch):
+    config_file = write_config(
+        audeer.path(tmpdir, "default.yaml"),
+        "audio:\n  activity_preroll_s: null\n",
+    )
+    monkeypatch.setenv("PKG_AUDIO__ACTIVITY_PREROLL_S", "0.5")
+    # ``types`` mirrors the nested structure of the configuration
+    config = audeer.load_configuration(
+        config_file,
+        env_prefix="PKG",
```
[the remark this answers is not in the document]
ChristianGeng: Added `test_load_configuration_environment_types_invalid_value` in 5434c4f: a non-float env value with `types={'audio': {'activity_preroll_s': float}}` raises `ValueError` with the 'could not be converted to the type' message.

ChristianGeng:
Addressed the review (commit 5434c4f); the five inline points are answered in their threads.

On the two high-level notes:

1. **Validate non-type `types` entries** — done. `_parse_environment_value` now guards `isinstance(target_type, type)` and raises a clear `ValueError` before reaching `issubclass()`.

2. **Make the nested `__` separator configurable** — declining this one. The `separator` argument is an *internal recursion parameter* of `_override_with_environment`, not part of the public `load_configuration()` API, so there is no caller-facing inconsistency to reconcile. The `__` nesting delimiter is a deliberate fixed convention (matching pydantic-settings' `env_nested_delimiter`), and the top-level `_` is fixed by the existing behaviour from #195. Exposing a knob would add public API surface with no current use case (YAGNI). Happy to revisit if a concrete need appears.

audeer/core/config.py
```diff
@@ -22,6 +23,14 @@ def load_configuration(
...
        applied in the given order
        (a later file overrides an earlier one)
+
+    Configuration files are deep-merged:
+    nested mappings are merged key by key,
+    so a section in a user file
+    only overrides the keys it defines
+    and keeps the remaining keys from the default.
+    Any non-mapping value (including a list)
+    replaces the previous value as a whole.
```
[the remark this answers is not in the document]
ChristianGeng: Fixed in 6904bca. The deep-merge paragraph was moved out of the numbered list to below item 3, so the 1/2/3 enumerated list stays contiguous and renders correctly in reStructuredText.

audeer/core/config.py
```diff
@@ -91,17 +120,42 @@ def load_configuration(
         if isinstance(user_config_files, str):
             user_config_files = [user_config_files]
         for user_config_file in user_config_files:
-            cfg.update(_load_configuration_file(user_config_file))
+            _deep_merge(cfg, _load_configuration_file(user_config_file))
 
     if env_prefix is not None:
-        _override_with_environment(cfg, env_prefix)
+        _override_with_environment(cfg, env_prefix, types=types or {})
```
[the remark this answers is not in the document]
ChristianGeng: Fixed in 6904bca. `load_configuration()` now raises a clear `ValueError` when `types` is not a mapping, instead of failing later with an `AttributeError` on `.get()`. Covered by `test_load_configuration_types_not_a_mapping`.

audeer/core/config.py
```diff
@@ -145,42 +199,80 @@ def _load_configuration_file(config_file: str) -> dict:
...
+    """
     for key, default_value in cfg.items():
-        name = f"{env_prefix}_{key.upper()}"
-        if name in os.environ:
+        name = f"{env_prefix}{separator}{key.upper()}"
+        key_type = types.get(key)
+        if isinstance(default_value, Mapping):
+            nested_types = key_type if isinstance(key_type, Mapping) else {}
+            _override_with_environment(default_value, name, nested_types, "__")
+        elif name in os.environ:
```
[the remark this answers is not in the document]
ChristianGeng: Fixed in 6904bca. A `types` entry for a nested section that is not itself a mapping (e.g. `types={'audio': float}`) now raises a clear `ValueError` instead of being silently ignored. Covered by `test_load_configuration_types_section_not_a_mapping`.

hagenw:
Regarding the three points you address in this pull request:

1. Makes sense (and point 2 is closely related, but in principle not required if 1. is restricted to override by user config files)
2. `PKG_MODEL__DEVICE`, why is this needed. The `PKG` refers to the package name, the `MODEL__DEVICE` part refers to the variable name, which should not have a `.` in its name. Or is the idea to support setting certain key-value paris of dictionary entries? E.g. let's assume we have `PKG_MODEL = '{"device": "cuda", "id": "abc"}]'`. With your addition, we could write `PKG_MODEL__DEVICE = "cpu"`. Could you provide an example for which this is needed. One downside is that it only works with dictionaries, but not with list of dictionaries as we have with `audb.config.REPOSITORIES`.
3. Can you provide an example for this. I guess you target a setting which has `None` in its system wide config file?

ChristianGeng:
> Regarding the three points you address in this pull request:
> 
> 1. Makes sense (and point 2 is closely related, but in principle not required if 1. is restricted to override by user config files)
> 2. `PKG_MODEL__DEVICE`, why is this needed. The `PKG` refers to the package name, the `MODEL__DEVICE` part refers to the variable name, which should not have a `.` in its name. Or is the idea to support setting certain key-value paris of dictionary entries? E.g. let's assume we have `PKG_MODEL = '{"device": "cuda", "id": "abc"}]'`. With your addition, we could write `PKG_MODEL__DEVICE = "cpu"`. Could you provide an example for which this is needed. One downside is that it only works with dictionaries, but not with list of dictionaries as we have with `audb.config.REPOSITORIES`.
> 3. Can you provide an example for this. I guess you target a setting which has `None` in its system wide config file?

One use case is to use env variables that get slotted beneath the top level hierarchy.  
For example we might have a `device` for several models that we use.  ` PKG_MODEL_B__DEVICE=cuda`  versus ` PKG_MODEL_A__DEVICE=cpu`. The other limitation you're right about: this reaches keys inside dicts, not elements of a list of dicts like `audb.config.REPOSITORIES`;

hagenw:
> > Regarding the three points you address in this pull request:
> > 
> > 1. Makes sense (and point 2 is closely related, but in principle not required if 1. is restricted to override by user config files)
> > 2. `PKG_MODEL__DEVICE`, why is this needed. The `PKG` refers to the package name, the `MODEL__DEVICE` part refers to the variable name, which should not have a `.` in its name. Or is the idea to support setting certain key-value paris of dictionary entries? E.g. let's assume we have `PKG_MODEL = '{"device": "cuda", "id": "abc"}]'`. With your addition, we could write `PKG_MODEL__DEVICE = "cpu"`. Could you provide an example for which this is needed. One downside is that it only works with dictionaries, but not with list of dictionaries as we have with `audb.config.REPOSITORIES`.
> > 3. Can you provide an example for this. I guess you target a setting which has `None` in its system wide config file?
> 
> One use case is to use env variables that get slotted beneath the top level hierarchy. For example we might have a `device` for several models that we use. ` PKG_MODEL_B__DEVICE=cuda` versus ` PKG_MODEL_A__DEVICE=cpu`. The other limitation you're right about: this reaches keys inside dicts, not elements of a list of dicts like `audb.config.REPOSITORIES`;

Ok, this makes sense. An alternative would be to use `PKG_MODEL_A = '{"device": "cuda"}'` and then to make sure it only overrides the matching part of the underlying dictionary. But `PKG_MODEL_A__DEVICE=cuda` reads easier to me. Is there an limit on the nesting or can you do `PKG_VAR__KEYA__KEYB__KEYC=value`?

ChristianGeng:
No limit — it recurses to arbitrary depth. `PKG_VAR__KEYA__KEYB__KEYC=value` maps to `cfg["var"]["keya"]["keyb"]["keyc"]`. Verified with 3 levels: `PKG_A__B__C=…` plus `PKG_A__B__D=42` yields `{"a": {"b": {"c": …, "d": 42}}}`, with `d` still cast to `int` from its default.

The only bounds are the documented ones: each intermediate level must already exist in the merged config (only existing keys can be overridden), and descent stops at the first non-mapping (leaf) value — which is also why it can't index into a list of dicts.

I am not sure whether one should limit recursion depth. I mean a depth of 10 is not so good I take it.

audeer/core/config.py
```diff
@@ -1,3 +1,293 @@
...
+            or a ``types`` entry for a nested section,
+            is not a mapping
+
+    Examples:
+        >>> import tempfile
+        >>> config_file = audeer.path(tempfile.mkdtemp(), "config.yaml")
+        >>> with open(config_file, "w") as file:
+        ...     _ = file.write("cache_root: ~/cache\n")
+        >>> audeer.load_configuration(config_file)
+        {'cache_root': '~/cache'}
```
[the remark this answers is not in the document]
ChristianGeng: The doctests run under Sybil with `audeer` injected into the namespace (see `audeer/core/conftest.py`: `namespace["audeer"] = audeer`), so this example executes and passes in CI without an explicit `import audeer`, consistent with the other examples in the module. Leaving it as is.

audeer/core/config.py
```diff
@@ -1,3 +1,293 @@
...
+        # any value other than the truthy ones becomes ``False``
+        if issubclass(target_type, bool):
+            return value.lower() in ("1", "true", "yes", "on")
+        if issubclass(target_type, int):
+            return int(value)
+        if issubclass(target_type, float):
+            return float(value)
+        # ``json.JSONDecodeError`` is a subclass of ``ValueError``
+        if issubclass(target_type, (list, dict)):
+            return json.loads(value)
```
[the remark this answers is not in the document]
ChristianGeng: Fixed in 917228b. `json.loads` accepts any JSON value, so for a `list`/`dict` target the parsed value is now checked with `isinstance(parsed, target_type)` and raises the standard wrapped `ValueError` otherwise, instead of silently returning the wrong type. Added `test_load_configuration_environment_types_json_wrong_type` (a `list` default with `PKG_HOSTS=123` now raises; a valid JSON list is still accepted).

audeer/core/config.py
```diff
@@ -59,6 +80,12 @@ def load_configuration(
         env_prefix: prefix of environment variables
             used to override configuration values.
             If ``None``, environment variables are ignored
+        types: mapping that declares the type
+            of configuration values,
+            mirroring the (possibly nested) configuration structure.
+            Used to cast environment variable overrides
+            for keys whose default value is ``None``,
+            or to override the type inferred from the default value
```
hagenw: Can you add an example to the "Examples" section, as for me it is completely clear how I have to provide it.

What do I have to provide if default is `None`, but when a value is given it should be a list. Should then be `list` or `string` (the representation of the list in the environment variable). Or do I not have to provide anything here as `string` is already the default as you mention in the docstring?

tests/test_config.py
```diff
@@ -127,6 +160,35 @@ def test_load_configuration_environment(tmpdir, monkeypatch):
...
+    # Nested keys are addressed with '__' between the levels,
+    # and the value is cast to the type of the nested default
+    monkeypatch.setenv("PKG_MODEL__DEVICE", "cpu")
+    monkeypatch.setenv("PKG_MODEL__LORA", "true")
+    monkeypatch.setenv("PKG_GENERATION__TOP_K", "40")
+    config = audeer.load_configuration(config_file, env_prefix="PKG")
+    assert config == {
+        "model": {"device": "cpu", "lora": True},
+        "generation": {"top_k": 40},
+    }
```
hagenw, line 209: Would be nice to also have a test here that leaves some of the values unchanged by environment variables.

---

Also what happens if we set the dict as JSON and with the nested entries, e.g.

```python
monkeypatch.setenv("PKG_MODEL", '{"device": "cuda", "lora": "true"}')
monkeypatch.setenv("PKG_MODEL__DEVICE", "cpu")
```
?

I would expect it first applies the variables higher in hierarchy (`PKG_MODEL`) and then `PKG_MODEL__DEVICE`, so we will end up having `model.device = "cpu"`.

tests/test_config.py
```diff
@@ -137,6 +199,132 @@ def test_load_configuration_environment_bool_false(tmpdir, monkeypatch):
...
+    config_file = write_config(
+        audeer.path(tmpdir, "default.yaml"),
+        "audio:\n  activity_preroll_s: null\n",
+    )
+    monkeypatch.setenv("PKG_AUDIO__ACTIVITY_PREROLL_S", "0.5")
+    # ``types`` mirrors the nested structure of the configuration
+    config = audeer.load_configuration(
+        config_file,
+        env_prefix="PKG",
+        types={"audio": {"activity_preroll_s": float}},
```
hagenw, line 509: If I specify `types={"audio": dict}` instead, will it return `{"audio": {"activity_preroll_s": "0.5"}}` or raise an error as below for `types={"audio": float}`?

hagenw:
> No limit — it recurses to arbitrary depth. `PKG_VAR__KEYA__KEYB__KEYC=value` maps to `cfg["var"]["keya"]["keyb"]["keyc"]`. Verified with 3 levels: `PKG_A__B__C=…` plus `PKG_A__B__D=42` yields `{"a": {"b": {"c": …, "d": 42}}}`, with `d` still cast to `int` from its default.
> 
> The only bounds are the documented ones: each intermediate level must already exist in the merged config (only existing keys can be overridden), and descent stops at the first non-mapping (leaf) value — which is also why it can't index into a list of dicts.
> 
> I am not sure whether one should limit recursion depth. I mean a depth of 10 is not so good I take it.

I think there is no need to limit the recursion depth. The user should be able to create meaningful config settings, so I would not add extra code that enforces limits here.

audeer/core/config.py
ChristianGeng: Apologies, you requested an example above already in point 3.  see fcf771b for one.

tests/test_config.py
ChristianGeng: With both `PKG_MODEL`&rsquo;{&ldquo;device&rdquo;: &ldquo;cuda&rdquo;, &ldquo;lora&rdquo;: &ldquo;true&rdquo;}&rsquo;= and `PKG_MODEL__DEVICE=cpu` set, the result is `model.device =` &ldquo;cpu&rdquo;=, as you expected.

But that comes only from `PKG_MODEL__DEVICE`. The whole-dict `PKG_MODEL` is never applied — entire-section overrides are unsupported, so `PKG_MODEL` on its own is a no-op (the section stays at the file default).

Should whole-dict overrides be implemented? If not, we should likely raise an error when a whole-section variable is set, instead of silently ignoring it. The test is not implemented - depends on what we decide here.

tests/test_config.py
ChristianGeng: Currently it raises, same as `types={"audio": float}`:

    ValueError: The 'types' entry for the nested section 'audio' must be a mapping, but is 'type'.

so it does not return `{"audio": {"activity_preroll_s": "0.5"}}`. A section&rsquo;s `types` entry has to be a mapping mirroring its sub-keys — to type the nested key you write:

```python
types={"audio": {"activity_preroll_s": float}}
```

so `PKG_AUDIO__ACTIVITY_PREROLL_S=0.5` resolves to `{"audio": {"activity_preroll_s": 0.5}}` (a float).

This is the same decision as the `PKG_MODEL` whole-dict thread: if we add whole-section JSON overrides, then `types={"audio": dict}` would be the natural way to declare &ldquo;parse this section as a JSON dict&rdquo;, and the answer flips from raise to parse. Until then it raises. So I would decide both together - I mean the thread above https://github.com/audeering/audeer/pull/196#discussion_r3630457492

audeer/core/config.py
```diff
@@ -84,24 +116,69 @@ def load_configuration(
         >>> audeer.load_configuration(config_file)
         {'cache_root': '~/cache'}
 
+        A key that defaults to ``None`` has no inferred type,
+        so declare it in ``types``.
+        The environment variable then holds
+        the JSON representation of the value,
+        e.g. a JSON array for a ``list``:
```
```suggestion
```

tests/test_config.py
hagenw: Good question if we should allow it or not.

Argument in favor of it:

* We support already the JSON syntax for lists, so a user expects that this works for arbitrary types

Argument against it:

* When the default value of `model` is `{"a": 0, "b": 1}` and a user sets `PKG_MODEL='{"a": 1}'` the expected outcome is not as obvious as for `PKG_MODEL__A=1`, should `model` contain `{"a": 1}` or `{"a": 1, "b": 1}` after specifying `PKG_MODEL`. For other types using the JSON syntax overrides the whole variable, so most likely we would expect `{"a": 1}`

---

Related question:

In the description we say `PKG_MODEL__DEVICE → model.device`. But what if we have `model["device"]` in the code. Should this then also be handled by `PKG_MODEL__DEVICE`? I'm asking because in the case of `a.b` and `a.c` it makes no sense to allow `PKG_A` to remove `c` from the entries, but for `a["b"]` and `a["c"]` it might be ok to replace the dictionary `a` with a complete new one.

hagenw:
Regarding my last comment discussing `a.b` vs. `a["b"]`. I think this should be irrelevant for our discussion here. We are anyway not changing the underlying code when providing configuration values. And if some values are not provided the code will still have default values.

So I would vote to support both approaches:

* `PKG_V='{"k1": "value"}'`
* `PKG_V__K1="value"`

and apply the following rules:

* `PKG_V` parses the JSON value and replaces the whole mapping
* Nested variables (e.g. `PKG_V__K1`) are applied afterwards and therefore have higher precedence

This means if we have

```yaml
v:
  k1: default
  k2: retained?
```
and
```bash
PKG_V='{"k1": "new"}'
```
we would get
```python
{"v": {"k1": "new"}}
```
whereas with
```bash
PKG_V__K1=new
```
we would get
```python
{"v": {"k1": "new", "k2": "retained?"}}
```

---

If I understand `pydantic-settings` [documentation on ENV parsing](https://pydantic.dev/docs/validation/latest/concepts/pydantic_settings/#parsing-environment-variable-values) correctly, it does the same.

hagenw:
I would stay with the rule that only existing config entries can be overridden by environment variables, but a dict entry like `PKG_MODEL='{"a": 1}'` is allowed to introduce `a` as we cannot know if this is desired or not. But if it is not desired, `"a"` is simply ignored in the code.

hagenw:
I also had a look at similar projects like `omegaconf` or `pydantic-settings`. `omegaconf` has not a big overlap with what we want to achieve here and both have additional dependencies, which makes it meaningful to include our own processing in `audeer`.

tests/test_config.py
hagenw: As discussed at https://github.com/audeering/audeer/pull/196#issuecomment-5057934806, I would vote for supporting entire section overrides with `PKG_MODEL`.

audeer/core/config.py
ChristianGeng: Applied via own commit. So can be resolved.

ChristianGeng:
> I also had a look at similar projects like `omegaconf` or `pydantic-settings`. `omegaconf` has not a big overlap with what we want to achieve here and both have additional dependencies, which makes it meaningful to include our own processing in `audeer`.

Sounds good, let us do it your way.

To make sure I understood. The single key setting can stay as is and is already implemented.
The new thing is the entire dict handling.  So

  `PKG_V='{"k1": "new"}'` replaces the whole `v` dict (NEW)

With this rule, a whole-dict replace does not need `types` (the JSON already says it is a dict. So it probably should continue to raise an error. 

 Should this go into this PR, or a separate follow-up?

hagenw:
> > I also had a look at similar projects like `omegaconf` or `pydantic-settings`. `omegaconf` has not a big overlap with what we want to achieve here and both have additional dependencies, which makes it meaningful to include our own processing in `audeer`.
> 
> Sounds good, let us do it your way.
> 
> To make sure I understood. The single key setting can stay as is and is already implemented. The new thing is the entire dict handling. So
> 
> `PKG_V='{"k1": "new"}'` replaces the whole `v` dict (NEW)
> 
> With this rule, a whole-dict replace does not need `types` (the JSON already says it is a dict. So it probably should continue to raise an error.

Yes, raising an error for `dict` is fine. Please add a test for it.
 
> Should this go into this PR, or a separate follow-up?

As this covers the core of this pull request, please implement it here.

Make sure to add tests for the following cases:

 - whole-dictionary JSON override;                                                                                                                                            
 - whole override followed by nested overrides;                                                                                                                               
 - invalid JSON;                                                                                                                                                              
 - JSON of the wrong shape, such as `PKG_V='[1, 2]'` for a mapping;                                                                                                             
 - replacement rather than deep-merge semantics;                                                                                                                              
 - a nested variable referring to a key omitted by the whole JSON value;                                                                                                      
 - behavior of unknown JSON keys.

ChristianGeng:
> > > I also had a look at similar projects like `omegaconf` or `pydantic-settings`. `omegaconf` has not a big overlap with what we want to achieve here and both have additional dependencies, which makes it meaningful to include our own processing in `audeer`.
> > 
> > 
> > Sounds good, let us do it your way.
> > To make sure I understood. The single key setting can stay as is and is already implemented. The new thing is the entire dict handling. So
> > `PKG_V='{"k1": "new"}'` replaces the whole `v` dict (NEW)
> > With this rule, a whole-dict replace does not need `types` (the JSON already says it is a dict. So it probably should continue to raise an error.
> 
> Yes, raising an error for `dict` is fine. Please add a test for it.
> 
> > Should this go into this PR, or a separate follow-up?
> 
> As this covers the core of this pull request, please implement it here.
> 
> Make sure to add tests for the following cases:
> 
> * whole-dictionary JSON override;
> * whole override followed by nested overrides;
> * invalid JSON;
> * JSON of the wrong shape, such as `PKG_V='[1, 2]'` for a mapping;
> * replacement rather than deep-merge semantics;
> * a nested variable referring to a key omitted by the whole JSON value;
> * behavior of unknown JSON keys.

Done in commits  6cb0211 and 46cf3bd

-  whole-dict replace,  nested applied afterwards (higher precedence), 
- key introduction, 
- `types={"section": dict}` still raises.

tests/test_config.py
hagenw: We decided to still raise an error here.

audeer/core/config.py
```diff
@@ -145,42 +229,100 @@ def _load_configuration_file(config_file: str) -> dict:
...
+            # as a JSON object; nested variables (e.g. PKG_MODEL__DEVICE) are
+            # applied afterwards and therefore take precedence.
+            if name in os.environ:
+                cfg[key] = _parse_environment_value(
+                    name,
+                    os.environ[name],
+                    default_value,
+                    dict,
+                )
+            _override_with_environment(cfg[key], name, key_type or {}, "__")
```
hagenw: If we want to stay with the `separator` argument, you need to change here to


```suggestion
            _override_with_environment(cfg[key], name, key_type or {}, f"{separator}{separator}")
```

audeer/core/config.py
```diff
@@ -145,42 +229,100 @@ def _load_configuration_file(config_file: str) -> dict:
...
+            f"is not a type: {target_type!r}."
+        )
     try:
-        # ``bool`` has to be checked before ``int``.
+        # ``bool`` has to be checked before ``int``,
+        # as ``bool`` is a subclass of ``int``.
         # A boolean never fails conversion:
         # any value other than the truthy ones becomes ``False``
-        if isinstance(default_value, bool):
+        if issubclass(target_type, bool):
```
hagenw: What is the idea here of changing to `issubclass`?

audeer/core/config.py
```diff
@@ -145,42 +229,100 @@ def _load_configuration_file(config_file: str) -> dict:
...
-        if isinstance(default_value, int):
+        if issubclass(target_type, int):
             return int(value)
-        if isinstance(default_value, float):
+        if issubclass(target_type, float):
             return float(value)
         # ``json.JSONDecodeError`` is a subclass of ``ValueError``
-        if isinstance(default_value, (list, dict)):
-            return json.loads(value)
+        if issubclass(target_type, (list, dict)):
```
hagenw: I guess we need to extend this here to any possible type not covered by the other cases. E.g. if you provide `PKG_INPUT_SIGNAL='numpy.zeros(10)` it is not covered by `list` or `dict`.

audeer/core/config.py
```diff
@@ -145,42 +229,100 @@ def _load_configuration_file(config_file: str) -> dict:
...
+            # A whole-section variable (e.g. PKG_MODEL) replaces the mapping
+            # as a JSON object; nested variables (e.g. PKG_MODEL__DEVICE) are
+            # applied afterwards and therefore take precedence.
+            if name in os.environ:
+                cfg[key] = _parse_environment_value(
+                    name,
+                    os.environ[name],
+                    default_value,
+                    dict,
+                )
```
hagenw, line 431: We need to preserve the default types when overriding a whole section:

After `PKG_MODEL` replaces a mapping, recursion infers nested types from the replacement rather than the merged file configuration. For a default `model.count: 1`, setting `PKG_MODEL='{"count": "one"}'` and `PKG_MODEL__COUNT=2` produces `{"count": "2"}` instead of integer `2`, contrary to the documented conversion using the corresponding default value. Preserve the original mapping as type metadata when applying subsequent nested overrides.

Please also add a test for it.

audeer/core/config.py
```diff
@@ -145,42 +229,100 @@ def _load_configuration_file(config_file: str) -> dict:
...
+        types: declared types mirroring ``cfg``,
+            used to cast values whose default is ``None``
+        separator: string joining ``env_prefix`` and the current key
+            (``"_"`` at the top level, ``"__"`` for nested levels)
+
+    """
     for key, default_value in cfg.items():
-        name = f"{env_prefix}_{key.upper()}"
-        if name in os.environ:
+        name = f"{env_prefix}{separator}{key.upper()}"
```
hagenw: Do not assume every nested mapping key is a string:

Recursive environment processing now calls `.upper()` on nested keys, so a valid YAML mapping such as `status: {200: ok}` raises `AttributeError` whenever `env_prefix` is provided, even if no matching environment variable exists. This regression affects dictionary-valued settings with numeric keys; non-string keys should either be skipped for environment matching or rejected with a documented validation error.

Please also add a test for it.

hagenw (commented):
Validate declared leaf types even when no environment variable is set:

A malformed declaration such as `types={"timeout": "float"}` is only checked when `PKG_TIMEOUT` exists. Without that environment variable, loading silently succeeds,         allowing configuration errors to remain hidden until deployment and contradicting the documented `ValueError` for a declared value that is not a class. Validate the complete applicable types structure before processing environment overrides.

tests/test_config.py
ChristianGeng: Yes — kept raising. Covered in 46cf3bd: `test_load_configuration_types_section_not_a_mapping` is now parametrized over both `float` and `dict`.

tests/test_config.py
```diff
@@ -57,6 +57,39 @@ def test_load_configuration_user_file(tmpdir):
...
+        "model:\n  lora: true\n",
+    )
+    # Nested sections are deep-merged:
+    # a key set only in the default section is kept,
+    # instead of the whole section being replaced
+    assert audeer.load_configuration(default_file, user_file) == {
+        "model": {"device": "cuda", "lora": True},
+    }
+
+
```
hagenw, line 78: Can you add a test for the case that the user config adds a new entry. With environment variables we allow to add a new key entry, so we should allow the same with user config files:

```suggestion
def test_load_configuration_deep_merge(tmpdir):
    default_file = write_config(
        audeer.path(tmpdir, "default.yaml"),
        "model:\n  device: cuda\n  lora: false\n",
    )
    user_file = write_config(
        audeer.path(tmpdir, "user.yaml"),
        "model:\n  lora: true\n",
    )
    # Nested sections are deep-merged:
    # a key set only in the default section is kept,
    # instead of the whole section being replaced
    assert audeer.load_configuration(default_file, user_file) == {
        "model": {"device": "cuda", "lora": True},
    }


def test_load_configuration_deep_merge_user_adds_new_key(tmpdir):
    default_file = write_config(
        audeer.path(tmpdir, "default.yaml"),
        "model:\n  device: cuda\n  lora: false\n",
    )
    user_file = write_config(
        audeer.path(tmpdir, "user.yaml"),
        "model:\n  uid: abcdefg-1.0.0\n",
    )
    assert audeer.load_configuration(default_file, user_file) == {
        "model": {"device": "cuda", "lora": False, "uid": "abcdefg-1.0.0"},
    }

```

---

But how should we handle new entries on the main level? With environment variables this is not possible. But what should happen in the case of

```python
def test_load_configuration_deep_merge_user_adds_new_top_level(tmpdir):
   default_file = write_config(
        audeer.path(tmpdir, "default.yaml"),
        "model:\n  device: cuda\n  lora: false\n",
    )
    user_file = write_config(
        audeer.path(tmpdir, "user.yaml"),
        "tts:\n  vendor: iva-tts\n",
    )
    # Skip newly introduced top keys?
    assert audeer.load_configuration(default_file, user_file) == {
        "model": {"device": "cuda", "lora": False},
    }
```

tests/test_config.py
```diff
@@ -137,6 +280,136 @@ def test_load_configuration_environment_bool_false(tmpdir, monkeypatch):
...
+    # A declared type that is not a class
+    # raises a clear error instead of a TypeError from issubclass()
+    with pytest.raises(ValueError, match="is not a type"):
+        audeer.load_configuration(
+            config_file,
+            env_prefix="PKG",
+            types={"timeout": "float"},
+        )
+
+
```
hagenw, line 642: A test for the `types={"model": dict}` is missing, or have I overlooked it?

audeer/core/config.py
ChristianGeng: Fixed in e021d81 — non-string keys are now skipped in the env-matching loop (they can't be addressed by a variable name), so `status: {200: ok}` no longer crashes. Test added.

tests/test_config.py
hagenw: Sorry, I found now https://github.com/audeering/audeer/commit/46cf3bda6937d7c1bd09afcae92496d80334e003

audeer/core/config.py
ChristianGeng: `issubclass` is used so a `bool` default is still recognised as int-like: `bool` is a subclass of `int`, so it is checked first and cast correctly.

```python
>>> issubclass(bool, int)  # a bool IS an int
True
>>> bool is int            # but bool is not literally int
False
```

Testing `target_type is int` would miss a `bool`; `issubclass(target_type, int)` catches it.

hagenw: So the alternative would be to have `isinstance(var, bool) or isinstance(var, int)`, correct?

audeer/core/config.py
hagenw: > they can't be addressed by a variable name

But they can still be addressed by `PKG_STATUS='{200: "ok"}'`, or not?

Or should we forbid non-string keys in the configuration file? Do we allow for any valid YAML in the config at the moment, or do we have some restrictions already?

audeer/core/config.py
ChristianGeng: Fixed in 2b78acc. The original leaf types are captured before the JSON replaces the section and reused when casting the nested overrides, so `PKG_MODEL='{"count": "one"}'` + `PKG_MODEL__COUNT=2` now yields `{"count": 2}` (int). An explicit `types` entry still wins. Regression test added.

audeer/core/config.py
ChristianGeng: >> But they can still be addressed by PKG_STATUS='{200: "ok"}', or not?

Not this exact string.  200 would need quotattiion. 
 `PKG_STATUS='{"200": "ok2"}'` would work and replace  the whole section resulting in  ` {'status': {'200': 'ok2'}} `.
>> Or should we forbid non-string keys in the configuration file? 

I would probably not do that: They are valid yaml. 

>>  Do we allow for any valid YAML in the config at the moment, or do we have some restrictions already?

No currrenly we have no restrictions, any valid `yaml` is supported.  The only restriction is that the top level must be a `dict` / `mapping` type - but that might change.

tests/test_config.py
ChristianGeng: A user config file already adds keys at any level through the deep-merge, I would therefore allow additions at all levels. It is consistent with the nested case you already asked for (a user file adding `model.uid`), and extra keys the code does not read are harmless. So the test asserts the new top-level key is kept:

```python
def test_load_configuration_deep_merge_user_adds_new_top_level(tmpdir):
    default_file = write_config(
        audeer.path(tmpdir, "default.yaml"),
        "model:\n  device: cuda\n  lora: false\n",
    )
    user_file = write_config(
        audeer.path(tmpdir, "user.yaml"),
        "tts:\n  vendor: iva-tts\n",
    )
    assert audeer.load_configuration(default_file, user_file) == {
        "model": {"device": "cuda", "lora": False},
        "tts": {"vendor": "iva-tts"},
    }
```

audeer/core/config.py
ChristianGeng: Is it a good move to support arbitrary types here?
 Turning `numpy.zeros(10)` into an object from a string needs `eval` (or `audobject`?).
`eval`  is arbitrary code execution from an environment variable which is considered a security risk.
A plain constructor call would not help either, since `numpy` is not reconstructible from such a string.

So for a type outside `bool/int/float/list/dict` the value is returned as the raw string and the consuming code interprets it. Possibly this can be softened for types like `pathlib.Path`?

tests/test_config.py
hagenw: Ok, makes sense, please add the test.

And for your last example I assume is not possible with environment variable as `PKG_TTS=...` would fail as `tts` does not exist as possible config value. Which is fine.

audeer/core/config.py
hagenw: You are right. I first thought we need it for https://github.com/audeering/audb/issues/485 (which uses `audb.Repository`). But the trick is not to provide the object but the attributes, e.g.

```bash
AUDB_CONFIG_REPOSITORIES='[{"name": "audb-internal", "backend": "s3", "host": "s3.dualstack.eu-north-1.amazonaws.com"}]
```

So we can stay here with `list` and `dict`.

audeer/core/config.py
hagenw: Ok, let's stay with the limit for environment variables and do not restrict the yaml config files.

In the documentation we currently say:

> Environment variables are matched against the upper-cased configuration keys, prefixed with ``env_prefix`` and an underscore

I guess we should then update this to

> Environment variables are matched against upper-cased string-based configuration keys, prefixed with ``env_prefix`` and an underscore

? Or add another sentence state that it is limited to string keys.

ChristianGeng:
> Validate declared leaf types even when no environment variable is set:
> 
> A malformed declaration such as `types={"timeout": "float"}` is only checked when `PKG_TIMEOUT` exists. Without that environment variable, loading silently succeeds, allowing configuration errors to remain hidden until deployment and contradicting the documented `ValueError` for a declared value that is not a class. Validate the complete applicable types structure before processing environment overrides.

Fixed in 9eb2d85. Added a `_validate_types` pass over the applicable `types` structure that runs before the environment overrides, so a malformed declaration like `types={"timeout": "float"}` now raises even without `PKG_TIMEOUT` set. Regression test added.

tests/test_config.py
ChristianGeng: Added in 6af551d. `test_load_configuration_deep_merge_user_adds_new_top_level` asserts the new top-level key is kept. And agreed on the env-var point — `PKG_TTS` would not apply, since `tts` is not an existing config key, which is fine.

audeer/core/config.py
ChristianGeng: Done in a188dcd — added a sentence to the `load_configuration` docstring: only string keys are matched by environment overrides; a non-string key (e.g. a numeric YAML key) is left untouched.

audeer/core/config.py
ChristianGeng: We do not always have a value: [`target_type` may be declared](https://github.com/audeering/audeer/blob/a188dcdae7187ee6fc3973586048129ad7909749/audeer/core/config.py#L340-L341) via `types` (e.g. a `None` default typed as `float`), so we skip the inference and test the declared class directly with [`issubclass`](https://github.com/audeering/audeer/blob/a188dcdae7187ee6fc3973586048129ad7909749/audeer/core/config.py#L347-L354). 

So when  `var` is  a type we will do the wrong thing:

```python 
 >>> isinstance(float, float)   # asks: is the CLASS float a float VALUE?
False                          # no error — just False
>>> issubclass(float, float)
True
```

hagenw: Ok, then let's stay with the current solution

audeer/core/config.py
ChristianGeng: The nested delimiter is a fixed `__` at every level below the top, so it cannot be derived from `separator` by doubling. `f"{separator}{separator}"` gives `__` at depth 2 but `____` at depth 3, which drops the override.
a47741a contains a test for level 3.

hagenw: Ok, but we can also not remove `separator` as an argument as we need to set it to `__` when calling in certain cases? If not, please remove it. If we need it, we stay with the current implementation.

ChristianGeng: `separator` was internal to the private method, and no caller ever used it as the naming convention (`PKG_VAR` at the top level, `PKG_VAR__KEY` below) is fixed. 
Removed in 5cae917, the function does not have the argument any longer.

audeer/core/config.py
```diff
@@ -142,45 +233,136 @@ def _load_configuration_file(config_file: str) -> dict:
...
+            # as a JSON object; nested variables (e.g. PKG_MODEL__DEVICE) are
+            # applied afterwards and therefore take precedence.
+            nested_types = dict(key_type or {})
+            if name in os.environ:
+                # The JSON replacement loses the default types, so remember
+                # the original leaf types and reuse them when casting the
+                # nested overrides below (an explicit ``types`` entry wins).
+                for nested_key, nested_default in default_value.items():
+                    if not isinstance(nested_default, Mapping):
+                        nested_types.setdefault(nested_key, type(nested_default))
```
hagenw: Seems like we still have a bug here:

```python
import audeer
import os

with open("config.yaml", "w") as fp:
    fp.write("a:\n  b:\n    c: 1\n")

os.environ["PKG_A"] = '{"b":{"c":"bad"}}'
os.environ["PKG_A__B__C"] = "2"
audeer.load_configuration("config.yaml", env_prefix="PKG")
```

produces 

```
{'a': {'b': {'c': '2'}}}
```

but the desired output is

```
{'a': {'b': {'c': 2}}}
```

I think it should have raised an error for `PKG_A`?

audeer/core/config.py
```diff
@@ -142,45 +233,136 @@ def _load_configuration_file(config_file: str) -> dict:
...
-        if isinstance(default_value, (list, dict)):
-            return json.loads(value)
+        if issubclass(target_type, (list, dict)):
+            parsed = json.loads(value)
+            # ``json.loads`` accepts any JSON value,
+            # so a scalar like ``"123"`` parses without
+            # being the requested ``list``/``dict``.
+            if not isinstance(parsed, target_type):
+                raise ValueError
+            return parsed
```
hagenw: Running 

```python
import audeer
import os
import pathlib

with open("config.yaml", "w") as fp:
    fp.write("path: null\n")

os.environ["PKG_PATH"] = "file.txt"

audeer.load_configuration("config.yaml", env_prefix="PKG", types={"path": pathlib.Path})
```

returns

```
{'path': 'file.txt'}
```

silently ignoring the `pathlib.Path` type. Maybe we should raise a `ValueError` if an unsupported type is given instead of silently converting to string?

audeer/core/config.py
```diff
@@ -142,45 +233,136 @@ def _load_configuration_file(config_file: str) -> dict:
...
+        cfg: configuration dictionary
+        types: declared types mirroring ``cfg``
+
+    Raises:
+        ValueError: if a declared section type is not a mapping
+        ValueError: if a declared leaf type is not a class
+
+    """
+    for key, value in cfg.items():
+        declared = types.get(key)
```
hagenw: This avoids raising an error if we declare `None` as type.

```python
import audeer
import os

with open("config.yaml", "w") as file:
    file.write("timeout: null\n")

os.environ["PKG_TIMEOUT"] = "2.5"

audeer.load_configuration("config.yaml", env_prefix="PKG", types={"timeout": None})
```

returns

```
{'timeout': '2.5'}
```

but in line with the documentation it should have raised the following `ValueError` instead:

```
ValueError: The 'types' entry for 'timeout' is not a type: None.
```

audeer/core/config.py
ChristianGeng: Yes, we should. Fixed in 05a144c. A declared type outside the supported set now raises up
front, also when no environment variable is set:

```
ValueError: The 'types' entry for 'path' is not a supported type: 'Path'. Supported types are 'bool', 'int', 'float', 'str', 'list', 'dict'.
```

audeer/core/config.py
ChristianGeng: Fixed in 62882f7. The JSON object of a whole-section variable is now validated
recursively against the default values it replaces, 
Your example now raises

```
ValueError: The environment variable 'PKG_A={"b":{"c":"bad"}}' sets 'b.c' to a value of type 'str', but the corresponding configuration value has type 'int'.
```

and with a valid replacement such as `PKG_A='{"b":{"c":3}}'`, `PKG_A__B__C=2`
yields `{'a': {'b': {'c': 2}}}` with `c` an `int`.

audeer/core/config.py
ChristianGeng: An explicit `None` entry is no longer treated like an absent
key; your example now raises the documented
`ValueError: The 'types' entry for 'timeout' is not a type: None.`
Fixed in 527ad18.

audeer/core/config.py
```diff
@@ -142,45 +242,235 @@ def _load_configuration_file(config_file: str) -> dict:
...
+                raise ValueError(
+                    f"The 'types' entry for the nested section '{key}' "
+                    f"must be a mapping, but is '{type(declared).__name__}'."
+                )
+            _validate_types(value, declared)
+        elif not isinstance(declared, type):
+            raise ValueError(
+                f"The 'types' entry for '{key}' is not a type: {declared!r}."
+            )
+        elif not issubclass(declared, (bool, int, float, str, list, dict)):
```
hagenw: The use of `subclass` means we can do the following:

```python
import os
import audeer

with open("config.yaml", "w") as fp:
    fp.write("value: null\n")

class CustomInt(int):
    pass

os.environ["PKG_VALUE"] = "1"
config = audeer.load_configuration("config.yaml", env_prefix="PKG", types={"value": CustomInt})
type(config["value"])
```

returns

```
int
```

not

```
CustomInt
```

so we should change here to


```suggestion
        elif declared not in (bool, int, float, str, list, dict):
```

then we would get

```
ValueError: The 'types' entry for 'value' is not a supported type: 'CustomInt'. Supported types are 'bool', 'int', 'float', 'str', 'list', 'dict'.
```

audeer/core/config.py
```diff
@@ -142,45 +242,235 @@ def _load_configuration_file(config_file: str) -> dict:
...
+    Args:
+        cfg: configuration dictionary
+        types: declared types mirroring ``cfg``
+
+    Raises:
+        ValueError: if a declared section type is not a mapping
+        ValueError: if a declared leaf type is not a class
+
+    """
+    for key, value in cfg.items():
```
hagenw, line 277: This only validates keys from the config file, which means we do not catch misspellings in the type definitions.

```python
import os
import audeer

with open("config.yaml", "w") as fp:
    fp.write("timeout: null\n")

os.environ["PKG_TIMEOUT"] = "2"
audeer.load_configuration("config.yaml", env_prefix="PKG", types={"tiemout": float})
```
returns

```
{'timeout': '2'}
```

Maybe we should instead raise an error that `tiemout` is not a supported entry?

Of cause it gets more complicated as we are allowed to add new entries with environment variables, e.g.

```python
import os
import audeer

with open("config.yaml", "w") as fp:
    fp.write("connection:\n  timeout: null\n")

os.environ["PKG_CONNECTION"] = '{"tiemout": 2}'
audeer.load_configuration("config.yaml", env_prefix="PKG", types={"connection": {"tiemout": float}})
```
returns
```
{'connection': {'tiemout': 2.0}}
```

But it feels safer if we raise an error if `types` has some non-matching key entry.

audeer/core/config.py
ChristianGeng: Fixed in 2871fa7. matching against declared type.

audeer/core/config.py
ChristianGeng: Fixed in 2190a61. `types` entries are now checked recursively against the configuration keys, so `types={"tiemout": float}` raises `ValueError: The 'types' entry 'tiemout' does not match any configuration key.` As you anticipated, this means a type can no longer be declared for a key that only enters via a JSON environment override (your second example now raises too) — the JSON value already carries its type.

hagenw: Could you add a test for the second example as well.

audeer/core/config.py
```diff
@@ -42,8 +71,25 @@ def load_configuration(
...
+
+    A default value of ``None`` carries no type,
+    so an environment variable would be kept as a string.
+    Use ``types`` to declare the intended type
+    for such keys,
+    or to override the inferred type of any key.
+    ``types`` mirrors the (possibly nested) structure
+    of the configuration;
+    an entry that does not match a configuration key
+    raises a ``ValueError`` to catch misspellings.
```
```suggestion
    an entry that does not match a configuration key
    raises a ``ValueError``.
```

audeer/core/config.py
ChristianGeng: Added in cedf690.

audeer/core/config.py
ChristianGeng: Applied in fe84396.

hagenw (approved):

PR #198, opened 2026-07-23
Whole-dict environment overrides (stacked on #196)

ChristianGeng:
Pre-staged follow-up implementing whole-dict environment overrides, discussed in #196.

**Base is the #196 branch** (`config-handling-helper-extras`), so it can be merged
into #196 if we decide to land the feature there; otherwise it stands alone as a
follow-up.

## What it adds

A whole nested mapping can be replaced by a single JSON variable, and nested
variables are layered on top:

- `PKG_MODEL='{"device": "cpu"}'` replaces the whole `model` mapping (keys it
  omits are dropped; it may introduce new keys).
- `PKG_MODEL__DEVICE=cpu` still overrides a single nested key.
- If both are set, the whole-dict variable is applied first and the nested one
  on top (higher precedence) — matching `pydantic-settings`.

Unchanged: only existing config entries are overridable (the whole-dict form is
the one exception, since it carries a full object); `types={"section": dict}`
still raises, since a JSON object is self-describing.

## Tests

Whole-dict replace, key introduction, whole-dict-then-nested precedence,
non-object rejection, and a partial-override case. `config.py` stays at 100%
coverage.

Refs #196.

### DIFF
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -43,6 +43,16 @@ def load_configuration(
     by joining the levels with ``__``,
     e.g. ``model.device``
     is overridden by ``<env_prefix>_MODEL__DEVICE``.
+    A whole nested mapping can instead be replaced
+    by a single variable holding a JSON object,
+    e.g. ``<env_prefix>_MODEL='{"device": "cpu"}'``;
+    the object replaces the mapping as a whole
+    (keys it omits are dropped)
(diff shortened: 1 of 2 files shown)


ChristianGeng:
Superseded — implemented directly in #196 (commits 6cb0211, 46cf3bd) per the review, so this stacked PR is no longer needed.

PR #199, opened 2026-07-28
Extend tests for load_configuration()

hagenw:
Extends the existing tests for `audeer.load_configuration()` by

| Test | Purpose |
| --- | --- |
| `test_load_configuration_missing_default_with_user_file` | missing default file with an existing user file |
| `test_load_configuration_deep_merge_scalar_replaces_section`, `test_load_configuration_deep_merge_section_replaces_scalar` | deep merge across the mapping/scalar boundary in both directions |
| `test_load_configuration_environment_bool_values` | parametrized over all documented truthy values (`"1"`, `"true"`, `"TRUE"`, `"Yes"`, `"on"`) |
| `test_load_configuration_environment_whole_dict_null_for_typed_default` | JSON `null` rejected against a `str` default and accepted for a `None` default |
| `test_load_configuration_environment_whole_dict_introduced_section_then_nested` | `PKG_MODEL__SUB__A` applies to a nested section that only exists because the whole-dict JSON introduced it |
| `test_load_configuration_validate_passes_with_environment` | validator success path, receiving the env-overridden config, with the config returned unchanged |
| `test_load_configuration_environment_empty_string` | `""` accepted for a `str` default, rejected for an `int` default |
| `test_load_configuration_environment_types_none_default_list` | the docstring's `types={"hosts": list}` example now covered by a unit test |

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -106,6 +106,34 @@ def test_load_configuration_deep_merge_user_adds_new_top_level(tmpdir):
     }
 
 
+def test_load_configuration_deep_merge_scalar_replaces_section(tmpdir):
+    default_file = write_config(
+        audeer.path(tmpdir, "default.yaml"),
+        "model:\n  device: cuda\n",
+    )
+    user_file = write_config(
+        audeer.path(tmpdir, "user.yaml"),
+        "model: none\n",
+    )
+    # A scalar in the user file replaces a whole default section
+    assert audeer.load_configuration(default_file, user_file) == {"model": "none"}
+
+
+def test_load_configuration_deep_merge_section_replaces_scalar(tmpdir):
+    default_file = write_config(
+        audeer.path(tmpdir, "default.yaml"),
+        "model: none\n",
+    )
+    user_file = write_config(
+        audeer.path(tmpdir, "user.yaml"),
+        "model:\n  device: cuda\n",
+    )
+    # A mapping in the user file replaces a scalar default as a whole
+    assert audeer.load_configuration(default_file, user_file) == {
+        "model": {"device": "cuda"},
+    }
+
+
 def test_load_configuration_deep_merge_list_replaced(tmpdir):
     default_file = write_config(
         audeer.path(tmpdir, "default.yaml"),
@@ -132,6 +160,19 @@ def test_load_configuration_missing_user_file(tmpdir):
     # the default configuration is kept
     assert audeer.load_configuration(default_file, user_file) == {
         "cache_root": "~/cache",
+    }
+
+
+def test_load_configuration_missing_default_with_user_file(tmpdir):
+    default_file = audeer.path(tmpdir, "missing.yaml")
+    user_file = write_config(
+        audeer.path(tmpdir, "user.yaml"),
+        "cache_root: ~/user\n",
+    )
+    # A missing default file yields an empty base,
+    # the user file provides the whole configuration
+    assert audeer.load_configuration(default_file, user_file) == {
+        "cache_root": "~/user",
     }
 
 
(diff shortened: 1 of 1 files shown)


tests/test_config.py
```diff
@@ -830,3 +971,23 @@ def validate(config):
...
+        audeer.path(tmpdir, "default.yaml"),
+        "count: 1\n",
+    )
+    monkeypatch.setenv("PKG_COUNT", "5")
+
+    calls = []
+    config = audeer.load_configuration(
+        config_file,
+        env_prefix="PKG",
+        validate=calls.append,
```
ChristianGeng: I think it should be like this:
```python

    def validate(config):
        # Snapshot: load_configuration() mutates and returns this very object
        calls.append(dict(config))

    config = audeer.load_configuration(
        config_file,
        env_prefix="PKG",
        validate=validate,
    )
```

Otehrwise Reference semantics bites: `calls.append(config)` binds a name to the same dict object, so tje `calls` list holds a reference to something that keeps changing.  So the inplace mutation is the dangerous bit.

hagenw: Good point, I updated accordingly.

tests/test_config.py
```diff
@@ -349,6 +390,32 @@ def test_load_configuration_environment_whole_dict_none_default(tmpdir, monkeypa
...
+    # JSON null does not match the str type of the default
+    monkeypatch.setenv("PKG_MODEL", '{"device": null}')
+    with pytest.raises(ValueError, match="has type 'str'"):
+        audeer.load_configuration(config_file, env_prefix="PKG")
+
+
+def test_load_configuration_environment_whole_dict_null_for_none_default(
+    tmpdir, monkeypatch
+):
+    config_file = write_config(
```
ChristianGeng, line 409: Should you give the assignment material to nullify, and add say a lora key:?

```python
    config_file = write_config(
        audeer.path(tmpdir, "default.yaml"),
        "model:\n  device: null\n  lora: false\n",
    )
```

Currently, it would pass anyway.

tests/test_config.py
```diff
@@ -463,6 +545,50 @@ def test_load_configuration_environment_bool_false(tmpdir, monkeypatch):
...
+        ("false", False),
+        ("", False),
+    ],
+)
+def test_load_configuration_environment_bool_values(
+    tmpdir, monkeypatch, value, expected
+):
+    config_file = write_config(
+        audeer.path(tmpdir, "default.yaml"),
+        "enabled: false\n",
```
ChristianGeng: The four falsy cases don't bite yet: the file already has `enabled: false`, so
their expected value equals the default. Turning `elif name in os.environ:` into
`elif False:` in `_override_with_environment()` leaves them green, while
`test_load_configuration_environment_bool_false` (default `enabled: true`) fails.

```suggestion
        f"enabled: {str(not expected).lower()}\n",
```

Then `bool_false` really is redundant and can join the list as `("no", False)`.

tests/test_config.py
hagenw: But that's not the purpose of the test here. It just tests that we are allowed to provide `null` in an environment variable and that this gets also converted to `None`. Adding `lora` would also test that this disappears because of overriding it with the environment variable. This is tested already at another place (I hope ;) ).

tests/test_config.py
hagenw: Changed accordingly.

ChristianGeng: Resolving.

tests/test_config.py
ChristianGeng: Ah I see, getting it now.  Then I am resolving.

ChristianGeng (approved):

Issue #197, opened 2026-07-22
load_configuration(): section-aware loading and effective-config provenance


ChristianGeng:
**Summary** — Two follow-ups to #196 (which stays at its current scope):

1. **Section-aware loading** — let `load_configuration` resolve a single
   caller-supplied section (simplest: accept a pre-parsed mapping, not only a
   file path), so several libraries can share one config file while each
   validates *its own* section against its own schema/defaults/prefix.
2. **Effective-config provenance** — after resolution, return which layer
   (default / file / env / explicit) set each key, so the caller can log the
   effective config. Independent of (1); could be its own issue.

## Context

Follow-up to #196 (deep-merge, nested env overrides, `types=`). Per the
discussion, #196 stays at its current scope; the two capabilities below are
proposed as separate, later work.

The motivating case: a top-level application composes settings from several
libraries, each of which owns its own schema and defaults. We want the libraries
to keep that ownership, while still resolving everything through one precedence
order (default < file < env < explicit) and, ideally, storing it in a single
config file. That raises two needs `load_configuration` does not cover yet.

## Part 1 — section-aware loading (each section validated by its owner)

A single file holds several independently-owned sections:

```yaml
lib-a:
  ...
lib-b:
  ...
app:
  ...
```

Each section should be resolvable against a *caller-supplied* schema, defaults,
`types`, and env prefix, so `lib-a` validates the `lib-a:` section rather than
the top-level app having to know every library's schema.

**Leading option (smallest change): accept a pre-parsed mapping as input.**
If `load_configuration` accepted an already-parsed `dict` (a single section)
in addition to a file path, the app could parse the one file itself and hand
each section to its owner:

```python
data = read_yaml("config.yaml")            # app parses once
cfg_a = lib_a.load_config(data["lib-a"])   # lib_a owns its schema/defaults/prefix
cfg_b = lib_b.load_config(data["lib-b"])
```

Each `lib_x.load_config` is a thin wrapper over `load_configuration(section,
defaults=..., types=..., env_prefix=...)`. This needs only a "accept a mapping,
not just a path" change — no native section syntax in the loader.

**Alternative: native named sections** inside `load_configuration` itself. More
convenient for callers, but a larger surface and it still has to delegate
per-section validation somehow. Raised for comparison; the pre-parsed-mapping
option is likely enough.

## Part 2 — effective-config provenance (separable, could be its own issue)

After precedence resolution it is hard to answer "what value is in effect for
[...]

PR #200, opened 2026-07-30
Accept mapping as user configuration

ChristianGeng:
Part 1 of #197 — the uncontroversial half, split out per hagenw's request in
[#197 (comment)](https://github.com/audeering/audeer/issues/197#issuecomment-5128236235).

`user_config_files` now accepts `str | Mapping | Sequence[str | Mapping]`
instead of only file paths, per the signature hagenw proposed in
[#197 (comment)](https://github.com/audeering/audeer/issues/197#issuecomment-5103562212).
This lets an application parse a configuration file shared by several
packages once and hand each section to the package that owns it:

```python
data = yaml.safe_load(open("config.yaml"))
cfg_a = lib_a.load_config(data["lib-a"])   # lib_a still owns its schema/defaults/prefix
cfg_b = lib_b.load_config(data["lib-b"])
```

`user_config_files` already accepted a bare `str` or a `Sequence[str]`; a
mapping follows the same bare-or-sequence shape, so it can stand alone or sit
next to file paths (and other mappings) in the sequence, rather than being a
special one-off case.

A mapping is deep-merged exactly like a file (same precedence, same
scalar/section rules) and is copied into plain dictionaries before merging,
so neither the merge nor later environment overrides can modify the mapping
given by the caller — including nested sections, lists, and tuples.

Part 2 (effective-config provenance) is intentionally not part of this PR;
still being discussed on the issue.

### DIFF
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -111,9 +111,8 @@ def load_configuration(
         default_config_file: path to default configuration file.
             The file does not have to exist
         user_configs: path(s) to user configuration file(s),
-            or an already parsed mapping,
-            applied in the given order
-            (a mapping may also appear as part of the sequence).
+            or already parsed mapping(s),
+            applied in the given order.
             Files do not have to exist,
             and a given mapping is not modified
         env_prefix: prefix of environment variables
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -215,11 +215,15 @@ def load_configuration(
 def _copy_mapping(mapping: Mapping) -> dict:
     r"""Copy a mapping into plain dictionaries.
 
-    Nested mappings and lists are copied as well,
+    Nested mappings, lists, and tuples are copied as well,
     so merging and environment overrides
     cannot modify the mapping given by the user.
-    All other values are immutable
-    or are never modified in place.
+    Values of other container types,
+    e.g. a ``set``,
+    are not copied and remain shared with the given mapping.
+    This is not a concern for configuration values
+    parsed from JSON or YAML,
+    which never produce such types.
 
     Args:
         mapping: mapping to copy
@@ -235,8 +239,8 @@ def _copy_value(value: object) -> object:
     r"""Copy a configuration value, see :func:`_copy_mapping`."""
     if isinstance(value, Mapping):
         return _copy_mapping(value)
-    if isinstance(value, list):
-        return [_copy_value(item) for item in value]
+    if isinstance(value, (list, tuple)):
+        return type(value)(_copy_value(item) for item in value)
     return value
 
 
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -7,7 +7,7 @@ import os
 
 def load_configuration(
     default_config_file: str,
-    user_config_files: str | Mapping | Sequence[str | Mapping] | None = None,
+    user_configs: str | Mapping | Sequence[str | Mapping] | None = None,
     *,
     env_prefix: str | None = None,
     types: Mapping | None = None,
@@ -20,7 +20,7 @@ def load_configuration(
 
     1. ``default_config_file``,
        e.g. the configuration file shipped with a package
-    2. ``user_config_files``,
+    2. ``user_configs``,
        applied in the given order
        (a later entry overrides an earlier one)
     3. environment variables,
@@ -110,7 +110,7 @@ def load_configuration(
     Args:
         default_config_file: path to default configuration file.
             The file does not have to exist
-        user_config_files: path(s) to user configuration file(s),
+        user_configs: path(s) to user configuration file(s),
             or an already parsed mapping,
             applied in the given order
             (a mapping may also appear as part of the sequence).
@@ -187,14 +187,14 @@ def load_configuration(
     """
     cfg = _load_configuration_file(default_config_file)
 
-    if user_config_files is not None:
-        if isinstance(user_config_files, (str, Mapping)):
-            user_config_files = [user_config_files]
-        for user_config_file in user_config_files:
-            if isinstance(user_config_file, Mapping):
-                _deep_merge(cfg, _copy_mapping(user_config_file))
+    if user_configs is not None:
+        if isinstance(user_configs, (str, Mapping)):
+            user_configs = [user_configs]
+        for user_config in user_configs:
+            if isinstance(user_config, Mapping):
+                _deep_merge(cfg, _copy_mapping(user_config))
             else:
-                _deep_merge(cfg, _load_configuration_file(user_config_file))
+                _deep_merge(cfg, _load_configuration_file(user_config))
 
     if types is not None:
         if not isinstance(types, Mapping):
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -192,9 +192,10 @@ def load_configuration(
             user_configs = [user_configs]
         for user_config in user_configs:
             if isinstance(user_config, Mapping):
-                _deep_merge(cfg, _copy_mapping(user_config))
+                update = _copy_mapping(user_config)
             else:
-                _deep_merge(cfg, _load_configuration_file(user_config))
+                update = _load_configuration_file(user_config)
+            _deep_merge(cfg, update)
 
     if types is not None:
         if not isinstance(types, Mapping):
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -85,7 +85,8 @@ def load_configuration(
     (e.g. a date parsed from YAML)
     cannot be overridden
     and raises a ``ValueError`` as well.
-    Only keys already present in the configuration files
+    Only keys already present in the merged configuration,
+    whether from a file or a mapping,
     can be overridden by environment variables.
     Only string keys are matched;
     a non-string key (e.g. a numeric YAML key)
@@ -110,12 +111,11 @@ def load_configuration(
         default_config_file: path to default configuration file.
             The file does not have to exist
         user_config_files: path(s) to user configuration file(s),
-            applied in the given order.
-            Files do not have to exist.
-            An already parsed mapping
-            can be given instead of a file path,
-            also as part of the sequence.
-            It is not modified
+            or an already parsed mapping,
+            applied in the given order
+            (a mapping may also appear as part of the sequence).
+            Files do not have to exist,
+            and a given mapping is not modified
         env_prefix: prefix of environment variables
             used to override configuration values.
             If ``None``, environment variables are ignored
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -7,7 +7,7 @@ import os
 
 def load_configuration(
     default_config_file: str,
-    user_config_files: str | Sequence[str] | None = None,
+    user_config_files: str | Mapping | Sequence[str | Mapping] | None = None,
     *,
     env_prefix: str | None = None,
     types: Mapping | None = None,
@@ -22,7 +22,7 @@ def load_configuration(
        e.g. the configuration file shipped with a package
     2. ``user_config_files``,
        applied in the given order
-       (a later file overrides an earlier one)
+       (a later entry overrides an earlier one)
     3. environment variables,
        if ``env_prefix`` is given
 
@@ -33,6 +33,16 @@ def load_configuration(
     and keeps the remaining keys from the default.
     Any non-mapping value (including a list)
     replaces the previous value as a whole.
+
+    A user configuration
+    may also be given as an already parsed mapping
+    instead of a file path,
+    e.g. a single section
+    of a configuration file
+    that the application has parsed itself
+    and that is shared by several packages.
+    It is deep-merged in the same way as a file,
+    and the given mapping is not modified.
 
     Environment variables are matched
     against the upper-cased configuration keys,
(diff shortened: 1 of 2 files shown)


audeer/core/config.py
```diff
@@ -34,6 +34,16 @@ def load_configuration(
...
 
+    A user configuration
+    may also be given as an already parsed mapping
+    instead of a file path,
+    e.g. a single section
+    of a configuration file
+    that the application has parsed itself
+    and that is shared by several packages.
+    It is deep-merged in the same way as a file,
+    and the given mapping is not modified.
```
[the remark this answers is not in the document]
ChristianGeng: Fixed in ddd0c58 — reworded the later sentence ("Only keys already present in the merged configuration, whether from a file or a mapping, can be overridden...") instead of adding a note here, so there's one source of truth rather than two docstring spots that could drift apart.

audeer/core/config.py
```diff
@@ -101,7 +111,11 @@ def load_configuration(
             The file does not have to exist
         user_config_files: path(s) to user configuration file(s),
             applied in the given order.
-            Files do not have to exist
+            Files do not have to exist.
+            An already parsed mapping
+            can be given instead of a file path,
+            also as part of the sequence.
+            It is not modified
```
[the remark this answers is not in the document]
ChristianGeng: Fixed in ddd0c58 — the Args entry now leads with the mapping option. Left out the trailing period though: none of the sibling Args entries (`default_config_file`, `env_prefix`, `types`, `validate`) end with one, so adding it here would be the odd one out rather than a fix.

ChristianGeng:
## Test Matrix Overview

@hagenw  Asked the AI to give a testing overview. Maybe this comes in handy for reviewing as these are many.

```text
Group A (1-6): does mapping merging work at all
Group B (7-10): is the caller's mapping protected from mutation
Group C (11-13): does it compose with env-override / types
Group D (14): does it work for any real Mapping, not just dict
Group E (15): does the end-to-end motivating use case work
```

| #  | Test                                                                                                                 | Checks                                                                                                                                                     | Kills / protects against                                                                                                                                                                                                                |
|--- |-------------------------------------------------------------------------------------------------------------------- |---------------------------------------------------------------------------------------------------------------------------------------------------------- |--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1  | [`test_load_configuration_user_mapping`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L201) | Bare mapping deep-merges into a nested section like a file would                                                                                           | Mapping support removed entirely                                                                                                                                                                                                        |
| 2  | [`..._after_file`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L214)                       | `[file, mapping]` — mapping is last, wins                                                                                                                  | Mapping support removed                                                                                                                                                                                                                 |
| 3  | [`..._before_file`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L231)                      | `[mapping, file]` — file is last, wins                                                                                                                     | Ordering not hardcoded to &ldquo;mapping always wins&rdquo;                                                                                                                                                                             |
| 4  | [`..._empty`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L247)                            | `{}` alone leaves defaults untouched                                                                                                                       | Weakest test: coincidentally passes even with mapping support fully broken                                                                                                                                                              |
| 5  | [`..._adds_new_key`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L258)                     | Mapping introduces keys absent from the default file, nested and top-level                                                                                 | Mapping support removed                                                                                                                                                                                                                 |
| 6  | [`..._list_replaced`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L272)                    | A list value from the mapping replaces the default list wholesale                                                                                          | Replace-not-merge rule for list values specifically                                                                                                                                                                                     |
| 7  | [`..._list_not_modified`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L283)                | Mutating the list (and a dict nested in it) after the call doesn&rsquo;t affect the result                                                                 | Sole killer of [list-copying removed](https://github.com/audeering/audeer/blob/ddd0c58/audeer/core/config.py#L242) in `_copy_value`                                                                                                     |
| 8  | [`..._tuple_not_modified`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L298)               | Same, for a tuple containing a dict                                                                                                                        | Sole killer of [tuple-copying removed](https://github.com/audeering/audeer/blob/ddd0c58/audeer/core/config.py#L242) — the bug sourcery-ai&rsquo;s review caught, fixed in [e12e1e3](https://github.com/audeering/audeer/commit/e12e1e3) |
| 9  | [`..._two_mappings`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L311)                     | Two mappings merged in sequence; merging the second doesn&rsquo;t mutate the first                                                                         | One of 4 tests catching &ldquo;no real copy happens&rdquo;                                                                                                                                                                              |
| 10 | [`..._not_modified`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L329)                     | Env-prefix override afterward doesn&rsquo;t touch the caller&rsquo;s original mapping, nor is the return value aliased to it                               | The general-purpose version of 7-9; same 4-test group                                                                                                                                                                                   |
| 11 | [`..._environment`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L353)                      | A mapping-introduced key is overridden by a plain env var, cast to the right type                                                                          | Mapping support removed                                                                                                                                                                                                                 |
| 12 | [`..._environment_json`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L374)                 | A mapping-introduced section is replaced by a JSON env var; wrong-typed JSON still raises                                                                  | Distinct code path from #11 — whole-section vs. scalar override                                                                                                                                                                         |
| 13 | [`..._types`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L391)                            | `types=` can declare a type for a key that only exists because the mapping introduced it                                                                   | Confirms `types` validation runs after merging, not before                                                                                                                                                                              |
| 14 | [`..._immutable`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L409)                        | A real `MappingProxyType` (nested, genuinely read-only) works as input and still gets copied                                                               | Strongest of the copy-guarantee tests — would **crash**, not just fail, if copying were skipped                                                                                                                                         |
| 15 | [`..._shared_file`](https://github.com/audeering/audeer/blob/ddd0c58/tests/test_config.py#L429)                      | One parsed YAML dict, two independent `load_configuration()` calls each consuming their own section, neither call affecting the other or the original dict | The actual motivating use case, end-to-end; cross-call isolation not covered by any unit test                                                                                                                                           |

hagenw:
There is still an open sourcery discussion. Is this ready for review already?

audeer/core/config.py
```diff
@@ -186,6 +212,38 @@ def load_configuration(
     return cfg
 
 
+def _copy_mapping(mapping: Mapping) -> dict:
```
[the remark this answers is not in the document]
hagenw: I would also vote for this simplification.

audeer/core/config.py
```diff
@@ -7,7 +7,7 @@
 
 def load_configuration(
     default_config_file: str,
-    user_config_files: str | Sequence[str] | None = None,
+    user_config_files: str | Mapping | Sequence[str | Mapping] | None = None,
```
hagenw: To better reflect the updated behavior, I would vote to rename `user_config_files`.
`user_config_files_or_mappings` seems to complicated, so I would just go with `user_configs`.

ChristianGeng: Renamed to `user_configs` in 6fb48fe.

audeer/core/config.py
```diff
@@ -164,11 +187,14 @@ def load_configuration(
...
-        for user_config_file in user_config_files:
-            _deep_merge(cfg, _load_configuration_file(user_config_file))
+    if user_configs is not None:
+        if isinstance(user_configs, (str, Mapping)):
+            user_configs = [user_configs]
+        for user_config in user_configs:
+            if isinstance(user_config, Mapping):
+                _deep_merge(cfg, _copy_mapping(user_config))
+            else:
+                _deep_merge(cfg, _load_configuration_file(user_config))
```
```suggestion
            if isinstance(user_config, Mapping):
                update = _copy_mapping(user_config)
            else:
                update = _load_configuration_file(user_config)
            _deep_merge(cfg, update)
```

What is the motivation to use `_copy_mapping()` and not `copy.deepcopy()`?

ChristianGeng: Applied the `update` refactor in 2bcadc2, thanks.

On `copy.deepcopy()` vs `_copy_mapping()`: `copy.deepcopy()` can't handle `MappingProxyType` at all —

```pycon
>>> import copy
>>> from types import MappingProxyType
>>> copy.deepcopy(MappingProxyType({"a": 1}))
TypeError: cannot pickle 'mappingproxy' object
```

— which would break `test_load_configuration_user_mapping_immutable`, the test that specifically guarantees any `Mapping` (not just `dict`) is accepted and protected from mutation. `copy.deepcopy()` does handle tuples and sets better than our current code (worth keeping in mind if we ever want to support those), but the `MappingProxyType` failure is a hard blocker on its own, so I'd keep `_copy_mapping`/`_copy_value` as they are.

ChristianGeng: Done in 2bcadc2bde48737682c31c1db3d22a9044aeb1a2.

ChristianGeng:
> There is still an open sourcery discussion. Is this ready for review already?

Yes — both sourcery-originated items are addressed now:

- The original tuple-copying suggestion was fixed in `e12e1e3` (already resolved).
- Your follow-up question on `copy.deepcopy()` vs `_copy_mapping()` is answered in [this reply](https://github.com/audeering/audeer/pull/200#discussion_r3688786483), with the concrete reason (`copy.deepcopy()` can't handle `MappingProxyType`, which would break `test_load_configuration_user_mapping_immutable`).

Sourcery itself is a bot review, not a person — nothing there needs answering beyond making sure its underlying findings are actually addressed, which they are. CI is green and the PR is mergeable, so this should be ready for review from our side.

audeer/core/config.py
hagenw: I had a closer look and `copy.deepcopy()` has indeed a few shortcomings (like not supporting read-only mappings like `immutabledict`), so it makes indeed sense to stay with the custom `_copy_mapping`.

audeer/core/config.py
```diff
@@ -99,9 +110,12 @@ def load_configuration(
     Args:
         default_config_file: path to default configuration file.
             The file does not have to exist
-        user_config_files: path(s) to user configuration file(s),
-            applied in the given order.
-            Files do not have to exist
+        user_configs: path(s) to user configuration file(s),
+            or an already parsed mapping,
+            applied in the given order
+            (a mapping may also appear as part of the sequence).
```
```suggestion
        user_configs: path(s) to user configuration file(s),
            or already parsed mapping(s),
            applied in the given order.
```

ChristianGeng: Applied in 5a76223, thanks.

audeer/core/config.py
hagenw: As configuration files and mappings are handled in the same way, we can simplify here to


```suggestion
   Configuration files and mappings are deep-merged:
   nested mappings are merged key by key,
   so a section in a user file
   only overrides the keys it defines
   and keeps the remaining keys from the default.
   Any non-mapping value (including a list)
   replaces the previous value as a whole.
```

audeer/core/config.py
```diff
@@ -186,6 +212,38 @@ def load_configuration(
...
+    """
+    return {key: _copy_value(value) for key, value in mapping.items()}
+
+
+def _copy_value(value: object) -> object:
+    r"""Copy a configuration value, see :func:`_copy_mapping`."""
+    if isinstance(value, Mapping):
+        return _copy_mapping(value)
+    if isinstance(value, (list, tuple)):
+        return type(value)(_copy_value(item) for item in value)
```
hagenw: This will fail if a user provides a `NamedTuple` as value, e.g. `audeer.load_configuration(default_file, {"position": Point(x=1, y=2)})`. I would in general not encourage to do this, but we can treat same as any other types (like we state for sets already in the config) by changing here to


```suggestion
    if type(value) in (list, tuple):
        return type(value)(_copy_value(item) for item in value)
```

audeer/core/config.py
hagenw: Sorry this is not correct.

hagenw: I propose to replace 

```python
    Configuration files are deep-merged:
    nested mappings are merged key by key,
    so a section in a user file
    only overrides the keys it defines
    and keeps the remaining keys from the default.
    Any non-mapping value (including a list)
    replaces the previous value as a whole.

    A user configuration
    may also be given as an already parsed mapping
    instead of a file path,
    e.g. a single section
    of a configuration file
    that the application has parsed itself
    and that is shared by several packages.
    It is deep-merged in the same way as a file,
    and the given mapping is not modified.
```

with

```python
    Configuration files or mappings are deep-merged:
    nested mappings are merged key by key,
    so a section in a user file
    only overrides the keys it defines
    and keeps the remaining keys from the default.
    Any non-mapping value (including a list)
    replaces the previous value as a whole.
```

ChristianGeng: Applied in 92f4757, thanks.

audeer/core/config.py
ChristianGeng: Good catch — reproduced it: `Point.__new__() missing 1 required positional argument: 'y'`. Fixed in 92f4757, same exact-type pattern as `_validate_types`'s supported-type check. Added a test with a real NamedTuple to lock it in.

hagenw (approved):

Issue #201, opened 2026-07-30
load_configuration(): effective-config tracking


ChristianGeng:
Part 2 of #197. Part 1 (`Mapping` support for `user_configs`) is PR #200.
This writes up part 2 — effective-config tracking, i.e. for each key in the
merged config, which layer set its value — as a standalone implementation
plan, the same way part 1 went from discussion to a dedicated PR.

----- 

Implementation Plan Details:

<details>

## Agreed so far

- hagenw asks for a concrete precedent before committing to the feature:
  ["sounds a little bit complicated to implement this. Do you have an
  example of another library doing this?"](https://github.com/audeering/audeer/issues/197#issuecomment-5103634846)
- Prior art posted: Dynaconf's `get_history()` / `inspect_settings()`, plus a
  local spike (branch
  [`provenance-tier0-spike`](https://github.com/audeering/audeer/tree/provenance-tier0-spike))
  showing the mechanism holds up against `load_configuration()` itself.
  [Reply](https://github.com/audeering/audeer/issues/197#issuecomment-5131025466)
- hagenw agrees to add the feature, but rules out a second return value, and
  lists three API shapes — a second function
  (`load_configuration_with_tracking()`), a hidden attribute plus an
  `inspect_configuration()` function, or an output parameter
  (`tracking={}`, citing `scipy.io.loadmat(mdict=...)` as precedent).
  [Comment](https://github.com/audeering/audeer/issues/197#issuecomment-5131550153)
- The output-parameter option is confirmed as the direction.
  [Reply](https://github.com/audeering/audeer/issues/197#issuecomment-5132541278)

## API

```python
def load_configuration(
    default_config_file: str,
    user_configs: str | Mapping | Sequence[str | Mapping] | None = None,
    *,
    env_prefix: str | None = None,
    types: Mapping | None = None,
    validate: Callable[[dict], None] | None = None,
    tracking: dict | None = None,
) -> dict:
```

- The return type does not change. `load_configuration()` always returns
  just the config `dict`, exactly as today. `tracking` is never returned —
  it works by mutating the dict passed in.
- Default `tracking=None` turns tracking fully off: no per-key bookkeeping
  runs at all. This is a structural no-op, not "compute it and discard it" —
  that would cost every existing caller something for a feature they never
  asked for.
- When a dict is passed, `load_configuration()` adds and updates its own
  entries in it, exactly like `dict.update()` — it never clears the dict
[...]

PR #202, opened 2026-07-31
Effective-config tracking for load_configuration()

ChristianGeng:
Relates to #201

Adds `tracking` to `load_configuration()`: a keyword-only mutable mapping that records which layer set each configuration key's effective value.

```python
tracking = {}
cfg = audeer.load_configuration(default_config_file, user_configs, env_prefix="PKG", tracking=tracking)
```

### Changes

- Add `tracking: MutableMapping | None = None` to `load_configuration()`
- Raise `ValueError` for a `tracking` value that is neither a mutable mapping nor `None`, matching `types=`'s own validation — this was a bug in the initial commit (silently treated any garbage value as "off" instead of raising), fixed in 9c38281
- Accept any `MutableMapping`, not only `dict` — consistent with `user_configs`/`types` already being typed as `Mapping` rather than `dict` — fixed in b324341
- Prefix `tracking` labels with `file:`/`env:` to match #201 and the spike (`"default"` and `mapping[<i>]` were already correct; file and environment-variable labels were missing their prefixes) — fixed in a70028d
- Stamp ownership directly inside `_override_with_environment()`, at the point it already holds the real key, instead of reconstructing a key from an environment variable's name (validated as ambiguous and crash-prone on a throwaway spike, [`provenance-tier0-spike`](https://github.com/audeering/audeer/tree/provenance-tier0-spike))

`tracking=None` (the default) is a genuine no-op, costing existing callers nothing. Passing a mapping adds entries into it in place, like `dict.update()` — reusing the same one across calls accumulates, matching the `scipy.io.loadmat(mdict=...)` pattern hagenw cited on #201.

Stacked on #200 (`user_configs`/`Mapping` support) — targets `user-config-mapping`, will need retargeting once #200 merges.

Verified: 614 tests passing, 100% coverage, ruff/pre-commit/Sphinx docs build clean.

### DIFF
--- a/audeer/core/config.py
+++ b/audeer/core/config.py
@@ -12,6 +12,7 @@ def load_configuration(
     env_prefix: str | None = None,
     types: Mapping | None = None,
     validate: Callable[[dict], None] | None = None,
+    tracking: dict | None = None,
 ) -> dict:
     r"""Load configuration from files and environment variables.
 
@@ -120,6 +121,39 @@ def load_configuration(
             dictionary and raises an error if it is invalid.
             It is applied once,
             after files and environment variables are merged
+        tracking: dict that records,
+            for each configuration key,
+            which layer set its effective value.
+            ``tracking`` mirrors the (possibly nested) structure
+            of the returned configuration,
+            and each leaf is a label naming its source:
+            ``"default"`` for ``default_config_file``,
+            the path of the ``user_configs`` entry
+            that provided a file,
+            ``"mapping[<i>]"`` for the ``user_configs`` entry
+            at index ``<i>`` that provided an already parsed mapping
+            (indices count every entry of the sequence,
+            file or mapping alike),
+            or the exact environment variable name
+            (e.g. ``"PKG_MODEL__DEVICE"``)
+            for an environment variable override.
+            A whole-section JSON environment variable
+            labels every key it sets;
+            a more specific nested variable applied afterwards
+            re-labels only the key it overrides,
+            leaving its siblings labeled by the section variable.
+            Entries are added to ``tracking`` in place,
+            exactly like ``dict.update()``;
+            it is never cleared first.
+            Pass a fresh ``{}`` for a clean view of a single call,
+            or reuse the same dict across several calls
+            to accumulate their entries.
+            ``tracking`` never influences the returned configuration,
+            it only records how it was assembled.
+            If ``None`` (the default),
+            no tracking is performed
+            and calling ``load_configuration()``
+            costs the same as without this argument
 
     Returns:
         merged configuration dictionary
@@ -173,18 +207,46 @@ def load_configuration(
         {'hosts': ['host1', 'host2']}
         >>> del os.environ["APP_HOSTS"]
 
+        ``tracking`` records which layer set each key's effective value.
+
+        >>> config_file = audeer.path(tempfile.mkdtemp(), "config.yaml")
+        >>> with open(config_file, "w") as file:
+        ...     _ = file.write("cache_root: ~/cache\n")
+        >>> tracking = {}
+        >>> audeer.load_configuration(
+        ...     config_file, {"cache_root": "~/user"}, tracking=tracking
+        ... )
+        {'cache_root': '~/user'}
+        >>> tracking
+        {'cache_root': 'mapping[0]'}
+
     """
     cfg = _load_configuration_file(default_config_file)
+
+    # Tracking is a genuine structural no-op when off: ``owner`` stays
+    # ``None``, so ``_deep_merge()``/``_override_with_environment()`` below
+    # run exactly the same path they always did, and no tracking tree is
+    # ever built. A non-``dict`` value (e.g. passed in error) is treated the
+    # same as ``None`` rather than crashing on an unrelated ``.update()``.
+    owner: dict | None = None
+    if isinstance(tracking, dict):
+        owner = _label_tree(cfg, "default")
 
     if user_configs is not None:
         if isinstance(user_configs, (str, Mapping)):
             user_configs = [user_configs]
-        for user_config in user_configs:
+        for i, user_config in enumerate(user_configs):
             if isinstance(user_config, Mapping):
                 update = _copy_mapping(user_config)
             else:
                 update = _load_configuration_file(user_config)
-            _deep_merge(cfg, update)
+            if owner is None:
+                _deep_merge(cfg, update)
+            else:
+                label = (
+                    f"mapping[{i}]" if isinstance(user_config, Mapping) else user_config
+                )
+                _deep_merge(cfg, update, owner, label)
 
     if types is not None:
         if not isinstance(types, Mapping):
@@ -194,10 +256,13 @@ def load_configuration(
         _validate_types(cfg, types)
 
     if env_prefix is not None:
-        _override_with_environment(cfg, f"{env_prefix}_", types or {})
+        _override_with_environment(cfg, f"{env_prefix}_", types or {}, owner)
 
     if validate is not None:
         validate(cfg)
+
+    if owner is not None:
+        tracking.update(owner)
 
     return cfg
 
@@ -237,7 +302,40 @@ def _copy_value(value: object) -> object:
     return value
 
 
-def _deep_merge(base: dict, update: dict) -> None:
+def _label_tree(mapping: Mapping, label: str) -> dict:
+    r"""Build a tracking tree that attributes every key to ``label``.
+
+    Mirrors the (possibly nested) structure of ``mapping``,
+    replacing every leaf value with ``label``.
+    Used both for the initial tracking tree
+    (every key starts out attributed to the default file)
+    and to expand a single label into a per-leaf tree,
+    e.g. when a whole configuration section
+    is replaced by a JSON environment variable
+    and a more specific, nested variable
+    later overrides one of its keys.
+
+    Args:
+        mapping: mapping whose structure is mirrored
+        label: label assigned to every leaf
+
+    Returns:
+        tracking tree, the same shape as ``mapping``,
+        with ``label`` at every leaf
+
+    """
+    return {
+        key: _label_tree(value, label) if isinstance(value, Mapping) else label
+        for key, value in mapping.items()
+    }
+
+
+def _deep_merge(
+    base: dict,
+    update: Mapping,
+    owner: dict | None = None,
+    label: str | None = None,
+) -> None:
     r"""Recursively merge ``update`` into ``base`` in place.
 
     Nested mappings are merged key by key,
@@ -246,9 +344,21 @@ def _deep_merge(base: dict, update: dict) -> None:
     Any non-mapping value (including lists)
     replaces the corresponding value in ``base``.
 
+    When ``owner`` is given,
+    it is updated in place to mirror ``base``:
+    a key that is merged key by key (recursion)
+    keeps or creates a nested dict in ``owner``,
+    and a key that is replaced wholesale
+    (a scalar, a list, or a mapping introduced fresh)
+    is attributed to ``label`` at every leaf of its subtree.
+
     Args:
         base: dictionary to merge into
         update: dictionary whose values take precedence
+        owner: tracking tree mirroring ``base``,
+            updated in place when given
+        label: label attributed to keys set or replaced by ``update``,
+            used only when ``owner`` is given
 
     """
     for key, value in update.items():
@@ -257,9 +367,20 @@ def _deep_merge(base: dict, update: dict) -> None:
             and isinstance(base[key], Mapping)
             and isinstance(value, Mapping)
         ):
-            _deep_merge(base[key], value)
+            if owner is None:
+                _deep_merge(base[key], value)
+            else:
+                # ``base[key]`` is a mapping, so ``owner[key]`` is
+                # already a matching nested dict: either from the
+                # initial tracking tree, or set by an earlier iteration
+                # of this same loop (see the ``else`` branch below)
+                _deep_merge(base[key], value, owner[key], label)
         else:
             base[key] = value
+            if owner is not None:
+                owner[key] = (
+                    _label_tree(value, label) if isinstance(value, Mapping) else label
+                )
 
 
 def _load_configuration_file(config_file: str) -> dict:
@@ -444,6 +565,7 @@ def _override_with_environment(
     cfg: dict,
     env_prefix: str,
     types: Mapping,
+    owner: dict | None = None,
 ) -> None:
     r"""Override configuration values with environment variables in place.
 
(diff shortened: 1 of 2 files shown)


ChristianGeng:
Both addressed.

> The `tracking` parameter type annotation and docstring currently imply
> only `dict | None` is valid...

Already fixed, before this review ran:
[c3157fd](https://github.com/audeering/audeer/commit/c3157fd)
raises `ValueError` for anything else, and
[5291935](https://github.com/audeering/audeer/commit/5291935)
widened the type hint to `MutableMapping | None`. This review ran against
the branch's original first commit, which predates both.

> When `owner` is provided to `_deep_merge`, `label` being `None` will
> propagate `None` into the tracking tree...

`_deep_merge` has exactly two call sites
([config.py:250-256](https://github.com/audeering/audeer/blob/61204f2/audeer/core/config.py#L250-L256)
and its own recursive call at
[config.py:384](https://github.com/audeering/audeer/blob/61204f2/audeer/core/config.py#L384)),
and both only pass `owner` together with a non-`None` `label`. So `owner`
set with `label=None` isn't reachable through the current call graph, not
just untested.

**Edit**: the branch was rebased onto `main` after #200 merged, so the
commit links above were re-pinned to their post-rebase SHAs.

audeer/core/config.py
```diff
@@ -120,6 +122,39 @@ def load_configuration(
...
             after files and environment variables are merged
+        tracking: mutable mapping that records,
+            for each configuration key,
+            which layer set its effective value.
+            ``tracking`` mirrors the (possibly nested) structure
+            of the returned configuration,
+            and each leaf is a label naming its source:
+            ``"default"`` for ``default_config_file``,
+            ``f"file:{path}"`` for the ``user_configs`` entry
+            that provided a file,
```
hagenw: I would propose to also return `file:{path}` instead of default, this makes it easier to find the actual configuration file.
As a compromise you could do `default:{path}`, but I think `tracking` is easier to understand if we limit the possible options to the three sources that are available: `file`, `mapping`, `env`.

audeer/core/config.py
```diff
@@ -246,9 +372,21 @@ def _deep_merge(base: dict, update: dict) -> None:
     Any non-mapping value (including lists)
     replaces the corresponding value in ``base``.
 
+    When ``owner`` is given,
+    it is updated in place to mirror ``base``:
+    a key that is merged key by key (recursion)
+    keeps or creates a nested dict in ``owner``,
+    and a key that is replaced wholesale
+    (a scalar, a list, or a mapping introduced fresh)
+    is attributed to ``label`` at every leaf of its subtree.
```
```suggestion
    When ``owner`` is given,
    it is updated in place like ``base``.
```

audeer/core/config.py
```diff
@@ -246,9 +372,21 @@ def _deep_merge(base: dict, update: dict) -> None:
...
+
     Args:
         base: dictionary to merge into
         update: dictionary whose values take precedence
+        owner: tracking tree mirroring ``base``,
+            updated in place when given
```
```suggestion
        owner: owner tracking dictionary to merge into
```

audeer/core/config.py
```diff
@@ -246,9 +372,21 @@ def _deep_merge(base: dict, update: dict) -> None:
...
+        label: label attributed to keys set or replaced by ``update``,
+            used only when ``owner`` is given
```
hagenw: Is it used only when `owner` is given, or is it required when `owner` is given? If the later, what happens if it is not provided?

audeer/core/config.py
```diff
@@ -257,9 +395,20 @@ def _deep_merge(base: dict, update: dict) -> None:
...
             and isinstance(value, Mapping)
         ):
-            _deep_merge(base[key], value)
+            if owner is None:
+                _deep_merge(base[key], value)
+            else:
+                # ``base[key]`` is a mapping, so ``owner[key]`` is
+                # already a matching nested dict: either from the
+                # initial tracking tree, or set by an earlier iteration
+                # of this same loop (see the ``else`` branch below)
```
```suggestion
```

audeer/core/config.py
hagenw: Also how is the mapping to the keys done?

Looking at the code it is called with:

```python
_deep_merge(base[key], value, owner[key], label)
```

it hands over the same labels for every key, which makes it hard to understand what will happen, e.g. is it a bug that we hand over the same labels, or is `label` a long list and we use different entries in each call?

Would it maybe make more sense to be more symmetrical here, e.g.

```python
def _deep_merge(
    base: dict,
    update: Mapping,
    owner: dict | None = None,
    owner_update: Mapping | None = None,
) -> None:
```
?

audeer/core/config.py
```diff
@@ -453,13 +603,30 @@ def _override_with_environment(
...
     is overridden by ``PKG_MODEL__DEVICE``.
 
+    When ``owner`` is given,
+    it is updated in place to mirror ``cfg``:
+    a key overridden by an environment variable
+    is attributed to that variable's exact name.
+    The key is already known at this point,
+    since it is what the variable name is built from,
+    so no attribution is ever derived
+    by reversing a variable name back into a key.
```
```suggestion
```

audeer/core/config.py
```diff
@@ -453,13 +603,30 @@ def _override_with_environment(
...
+
     Args:
         cfg: configuration dictionary, modified in place
         env_prefix: name prefix accumulated so far,
             including the trailing separator
             (``PKG_`` at the top level, ``PKG_MODEL__`` below)
         types: declared types mirroring ``cfg``,
             used to cast values whose default is ``None``
+        owner: tracking tree mirroring ``cfg``,
+            updated in place when given
```
```suggestion
        owner: owner tracking dictionary,
            updated in place when given
```

audeer/core/config.py
ChristianGeng: Done in 41cb349. Reduced to three source kinds — file, mapping, env — with the default file labeled `file:{path}` like any other file.

audeer/core/config.py
ChristianGeng: Done in a3c51d0.

audeer/core/config.py
ChristianGeng: Done in a3c51d0.

audeer/core/config.py
ChristianGeng: Required whenever `owner` is given — clarified in a3c51d0.

audeer/core/config.py
ChristianGeng: Done in a3c51d0.

audeer/core/config.py
ChristianGeng: Done in a3c51d0.

audeer/core/config.py
ChristianGeng: Done in a3c51d0 — trimmed the whole reverse-derivation justification sentence, not just this line, since the mechanical statement right above it already says what's needed and a partial cut would have left a dangling clause.

audeer/core/config.py
ChristianGeng: Every key `update` touches in one `_deep_merge()` call does get the same label intentionally, not a bug — they all come from the same `user_configs` entry (one file or one mapping), so there's nothing to differentiate per-key within that call. A per-key `owner_update: Mapping` mirroring `update` would end up with every leaf holding the identical label anyway, since the source is uniform; it adds a parallel structure without adding information. Clarified this directly in the `label` Args entry instead (a3c51d0) rather than restructuring the signature. Would the symmetric shape still be worth it for readability alone, even carrying no extra information per key?

hagenw: Thanks.

No, as we do not have symmetry here (as all come from the same source), you updated docstring is sufficient.

audeer/core/config.py
hagenw: Thanks.

audeer/core/config.py
```diff
@@ -120,6 +122,38 @@ def load_configuration(
...
+            it is never cleared first.
+            Pass a fresh ``{}`` for a clean view of a single call,
+            or reuse the same mapping across several calls
+            to accumulate their entries.
+            ``tracking`` never influences the returned configuration,
+            it only records how it was assembled.
+            If ``None`` (the default),
+            no tracking is performed
+            and calling ``load_configuration()``
+            costs the same as without this argument
```
hagenw: I think we can shorten here a lot ;)


```suggestion
        tracking: mutable mapping that records,
            for each configuration key,
            which layer set its effective value:
            ``f"file:{path}"`` for ``default_config_file``
            or a ``user_configs`` entry that provided a file,
            ``"mapping[<i>]"`` for the ``user_configs`` entry
            at index ``<i>`` that provided an already parsed mapping
            (indices count every entry of the sequence,
            file or mapping alike),
            or ``f"env:{name}"`` naming the exact environment variable
            (e.g. ``"env:PKG_MODEL__DEVICE"``)
            for an environment variable override.
            Entries are added to ``tracking`` in place.
            Reuse the same mapping across several calls
            to accumulate their entries.
            If ``None``,
            no tracking is performed
```

audeer/core/config.py
```diff
@@ -173,18 +207,73 @@ def load_configuration(
         {'hosts': ['host1', 'host2']}
         >>> del os.environ["APP_HOSTS"]
 
+        ``tracking`` records which layer set each key's effective value,
+        mirroring the (possibly nested) structure of the configuration.
+        A key untouched by any later layer keeps the default file's
+        own ``"file:"`` label, even a nested key sitting right next to
+        a sibling that an environment variable does override.
```
```suggestion
        ``tracking`` records which layer set each key's effective value.
```

ChristianGeng: Done in c74fbf9.

audeer/core/config.py
ChristianGeng: Mostly done in c74fbf9 — one addition: kept a trimmed one-sentence version of the whole-section-JSON-then-nested-override behavior. It's tested (`test_load_configuration_tracking_environment_whole_dict_then_nested`) but no example demonstrates it, so cutting it fully would have left it undocumented anywhere. Everything else went in as suggested.

audeer/core/config.py
hagenw, line 166:
```suggestion
       ValueError: if a ``types`` entry
           does not match any configuration key
       ValueError: if ``tracking`` is not a mutable mapping
```

audeer/core/config.py
```diff
@@ -173,18 +196,69 @@ def load_configuration(
...
+        >>> default_file_b = audeer.path(tempfile.mkdtemp(), "b.yaml")
+        >>> with open(default_file_b, "w") as file:
+        ...     _ = file.write("pool_size: 4\n")
+        >>> tracking = {}
+        >>> audeer.load_configuration(default_file_a, tracking=tracking)
+        {'cache_root': '~/cache'}
+        >>> audeer.load_configuration(default_file_b, tracking=tracking)
+        {'pool_size': 4}
+        >>> tracking
+        {'cache_root': 'file:...a.yaml', 'pool_size': 'file:...b.yaml'}
```
hagenw: I wonder if we are really interested in reusing the same tracking variable.

if we just reuse the same variable for different libraries, entries might be overridden, e.g.

```python
import audeer
import os
import tempfile

with tempfile.TemporaryDirectory() as config_root:
    # Two config files with top-level collision
    cfg_file1 = audeer.path(config_root, "pkg_a.yaml")
    cfg_file2 = audeer.path(config_root, "pkg_b.yaml")
    with open(cfg_file1, "w") as fp:
        fp.write("model:\n  device: cpu\n  lora: false\n")
    with open(cfg_file2, "w") as fp:
        fp.write("model:\n  sample_rate: 16000\n")
    
    os.environ["A_MODEL__DEVICE"] = "cuda"
    
    tracking = {}
    cfg1 = audeer.load_configuration(cfg_file1, env_prefix="A", tracking=tracking)
    cfg2 = audeer.load_configuration(cfg_file2, tracking=tracking)
```

Now, both package configs have different `model` settings (which makes sense as they might have different models):

```python
>>> cfg1
{'model': {'device': 'cuda', 'lora': False}}
>>> cfg2
{'model': {'sample_rate': 16000}}
```

But the corresponding tracking variable has only the entry for the second call, as `model` from the first call is overridden:

```python
>>> tracking
{'model': {'sample_rate': 'file:.../pkg_b.yaml'}}
```

In our use-case of having multiple entries in a shared config file, it also makes no sense to have a shared tracking variable. What we should do there is:

```python
data = yaml.safe_load(open("config.yaml"))
cfg_a = load_configuration(DEFAULTS_A, data.get("lib-a", {}), env_prefix="LIB_A", tracking=tracking_a)
cfg_b = load_configuration(DEFAULTS_B, data.get("lib-b", {}), env_prefix="LIB_B", tracking=tracking_b)
cfg = load_configuration(DEFAULTS, data.get("app", {}), env_prefix="APP", tracking=tracking)
```

So, I would propose to simply remove the example with sharing the tracking variable between two libraries.

audeer/core/config.py
```diff
@@ -491,19 +633,35 @@ def _override_with_environment(
...
                 # are applied on top of it as well
-                _override_with_environment(cfg[key], f"{name}__", {})
+                if owner is not None:
+                    owner[key] = _label_tree(cfg[key], f"env:{name}")
+                _override_with_environment(
+                    cfg[key],
+                    f"{name}__",
+                    {},
+                    owner[key] if owner is not None else None,
+                )
```
hagenw: I think the second part currently sets `owner[key]` again if we have a mapping, you can avoid this by:


```suggestion
            if isinstance(cfg[key], Mapping):
                # A mapping introduced via a declared ``dict`` type
                # behaves like a section, so nested variables
                # are applied on top of it as well
                _override_with_environment(cfg[key], f"{name}__", {})
                if owner is not None:
                    owner[key] = _label_tree(cfg[key], f"env:{name}")
                _override_with_environment(
                    cfg[key],
                    f"{name}__",
                    {},
                    owner[key] if owner is not None else None,
                )
            elif owner is not None:
                owner[key] = f"env:{name}"
```

audeer/core/config.py
ChristianGeng: Done in 890cf82.

audeer/core/config.py
ChristianGeng: Done in 3825d68 — removed the example. Reproduced your repro first to confirm it: tracking["model"] does end up with only pkg_b's entry, pkg_a's silently gone.

audeer/core/config.py
ChristianGeng: Done in 8a614a2, with one change: your suggestion calls `_override_with_environment()` twice on the same section (the extra unconditional call before the owner check) — applied the same idea (move the flat-label write into the non-Mapping branch) without that duplication. Verified behavior-preserving: same 631 tests pass, and the declared-dict-then-nested case gives byte-identical output.

hagenw: Yes, I had a typo in my suggestion. Glad that you found it.

hagenw (approved):

PR #206, opened 2026-08-04
CI: update checkout, setup-python, setup-uv

ChristianGeng:
setup-uv's cache defaults to keying on a `uv.lock`/`requirements*.txt` glob. `uv.lock` is gitignored in this repo (a library tests against a range of dependency versions, not one pinned set), so with the pinned `astral-sh/setup-uv@v5` the cache could never invalidate — every CI run printed:

```
No file matched to [**/uv.lock,**/requirements*.txt]. The cache will never get invalidated.
```

First attempt just disabled caching outright. Better fix, per @hagenw's question on whether caching could still help without a lockfile: bump the pinned action version instead. `v5`'s default `cache-dependency-glob` is literally `**/uv.lock` + `**/requirements*.txt` (confirmed from the `v5` tag's own `action.yml`) — neither exists here. [`v6.0.0`](https://github.com/astral-sh/setup-uv/releases/tag/v6.0.0) added `**/pyproject.toml` to the default glob specifically for this case. `pyproject.toml` is committed and changes exactly when a dependency actually changes, so bumping to a current version gives real caching — cache hit when nothing changed, correct invalidation when it does — with no lockfile required. Verified end to end on the actual CI run: first run reports a cache miss (expected, new key) and saves a fresh cache; the log shows the real glob used, including `pyproject.toml`.

Checked every breaking change from `v6` through `v9` against how these 4 workflows actually use the action (no `python-version`/`working-directory`/`manifest-file` passed to the `setup-uv` step, no self-hosted runners) — none apply.

Pinned to the exact `v9.0.0` tag, not a floating `v9` — confirmed via `git ls-remote --tags` that `astral-sh/setup-uv` doesn't publish a floating major-version tag (first push to this branch failed everything with "Unable to resolve action... unable to find version v9" before this was caught and fixed).

Side note, corrected: this does **not** fix the separate Node.js 20 deprecation warning. It removes `astral-sh/setup-uv` from that warning's list of flagged actions (that part of the fix genuinely is a `v7` change), but the warning itself still fires because `actions/checkout@v4` and `actions/setup-python@v5` are unrelated to this PR and still target Node 20.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -16,10 +16,10 @@ jobs:
         python-version: [ '3.12' ]
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v7
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v7
       with:
         python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -25,6 +25,8 @@ jobs:
 
     - name: Install uv
       uses: astral-sh/setup-uv@v9.0.0
+      with:
+        prune-cache: true
 
     - name: Test building documentation
       run: uv run python -m sphinx docs/ docs/_build/ -b html -W
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -25,6 +25,8 @@ jobs:
 
     - name: Install uv
       uses: astral-sh/setup-uv@v5
+      with:
+        enable-cache: false
 
     - name: Test building documentation
       run: uv run python -m sphinx docs/ docs/_build/ -b html -W
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -24,7 +24,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install uv
-      uses: astral-sh/setup-uv@v9
+      uses: astral-sh/setup-uv@v9.0.0
 
     - name: Test building documentation
       run: uv run python -m sphinx docs/ docs/_build/ -b html -W
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -25,8 +25,6 @@ jobs:
 
     - name: Install uv
       uses: astral-sh/setup-uv@v9.0.0
-      with:
-        prune-cache: true
 
     - name: Test building documentation
       run: uv run python -m sphinx docs/ docs/_build/ -b html -W
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -24,9 +24,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install uv
-      uses: astral-sh/setup-uv@v5
-      with:
-        enable-cache: false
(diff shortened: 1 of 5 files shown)


hagenw:
Ok, so the solution is to not cache at all as caching seems to work only if we commit the `uv.lock` file (which I'm not a fan of as we want to include always the latest packages in tests). I thought that it would still be faster with caching if we don't have new versions for some package, but you say this is not supported, correct?

ChristianGeng:
> Ok, so the solution is to not cache at all as caching seems to work only if we commit the `uv.lock` file (which I'm not a fan of as we want to include always the latest packages in tests). I thought that it would still be faster with caching if we don't have new versions for some package, but you say this is not supported, correct?

Yes, this is supported, I was wrong.

`@v5`'s pinned default `cache-dependency-glob` is literally `**/uv.lock` +
`**/requirements*.txt` (confirmed straight from the `v5` tag's `action.yml`),
neither of which exists here:

 [`v6.0.0`](https://github.com/astral-sh/setup-uv/releases/tag/v6.0.0)

This added `**/pyproject.toml` to the default glob specifically to fix this class
of complaint. `pyproject.toml` is committed and changes exactly when
dependencies change, so bumping the action version gives real caching:

- cache hit when nothing changed, 
- correct invalidation when it does

hagenw:
> [`v6.0.0`](https://github.com/astral-sh/setup-uv/releases/tag/v6.0.0)
> 
> This added `**/pyproject.toml` to the default glob specifically to fix this class of complaint. `pyproject.toml` is committed and changes exactly when dependencies change, so bumping the action version gives real caching:
> 
> * cache hit when nothing changed,
> * correct invalidation when it does

But this also means it will only check for new packages when we update `pyproject.toml`? Not sure if this is really what we want.

ChristianGeng:
> But this also means it will only check for new packages when we update pyproject.toml? Not sure if this is really what we want.

I think what happens is this: the cache only stores already-downloaded packages, not a decision about which version to use - uv sync still checks PyPI fresh every run regardless of cache state, so a new release gets picked up immediately either way. So it will pick up an upgrade. 

The only behavior change is that [`v9`](https://github.com/astral-sh/setup-uv/releases/tag/v9.0.0)
changed `prune-cache`'s default from `true` to `false`, so the cache
would otherwise grow unbounded instead of shrinking back down after
each run. 

I have set `prune-cache: true` explicitly to restore the old  default — pushed.
So we should now have the identical configuration but without the warnings.

hagenw:
Thanks, sounds fine.

When reading https://github.com/astral-sh/setup-uv/pull/967, it seems to me that `prune-cache: true` only makes sense for projects that download big pre-build packages like `torch`. As we don't have this in `audeer`, I think we can just go with the default setting.

ChristianGeng:
Good point, checked the actual benchmark data in [astral-sh/setup-uv#967](https://github.com/astral-sh/setup-uv/pull/967) — for a tiny dependency tree like audeer's (just `tqdm` + optional `pyyaml`), pruning saves 0 MB while costing 2 extra re-downloads per run that would otherwise be free cache hits. Undid [1d16748](https://github.com/audeering/audeer/pull/206/commits/1d16748a22178a4d4d67434f48e580cc639e1164) via revert in 0d57e11 — back to the plain `v9` default.

hagenw (approved):
Cool, thanks for solving the deprecation issues. Now we (Claude) just need to apply this to other repos as well.

PR #207, opened 2026-08-06
Let only the Test workflow save the shared uv cache

ChristianGeng:
## Problem

Every workflow in this repo installs uv via `astral-sh/setup-uv`,
and each one derives the same cache key — same runner OS, same Python version,
same dependency-file hash. When a single push triggers several workflows at
once, they all try to reserve that one key, and only the first to finish can
save it. The others log:

```
Failed to save: Unable to reserve cache with key setup-uv-2-x86_64-unknown-linux-gnu-ubuntu-24.04-3.12.13-5899ac3c8d62f4b9cac3f78a0d08699f23e7082de0032680922a422d775c3ad4, another job may be creating this cache.
```

Two live examples, both from the **same push to `main`** — note the identical
cache key in each, which is the collision itself:

| Workflow | Job | Run |
| --- | --- | --- |
| Documentation | `build (ubuntu-latest, 3.12)` | [30991884303](https://github.com/audeering/audeer/actions/runs/30991884303) |
| Linter | `build` | [30991884412](https://github.com/audeering/audeer/actions/runs/30991884412) |

Both failed to reserve, so a third workflow from that same push won the race.

The warning is harmless in itself — the losing job's cache content would have
been byte-identical to the winner's, and nothing fails. It is pure log noise,
but it appears on most pushes and obscures annotations that do matter.

## Fix

Keep the single shared cache key, but let only one workflow write it:
Documentation, Linter and Publish set `save-cache: false` on their `setup-uv`
step (the input is available in v9.0.0), while Test — the job with the fullest
environment — remains the sole saver. With one writer there is no reservation
race and no warning, cross-workflow cache sharing is preserved, and only a
single cache entry is stored.

## Trade-off

After a cache-key rotation (e.g. a `pyproject.toml` change), Documentation and
Linter miss the cache until the next Test run saves the new key — a one-run
cold start costing roughly a second or two per job.

## Context

This is a gap-closing change. This PR started as a per-workflow-cache draft
(`cache-suffix: ${{ github.workflow }}`) and pivoted to the single-saver
design after review feedback: per-workflow suffixes would have stored several
near-identical caches and given up cross-workflow sharing. `audeer` had
already merged audeering/audeer#206 before the warning was diagnosed, so it
never received a fix. The same gap exists in `opensmile-python`, `audformat`
and `audb`, and each will get the same single-saver pattern in a follow-up PR.

## Test plan

- [ ] CI green
- [ ] No `Unable to reserve cache` annotation on any job of this PR's runs

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -25,6 +25,8 @@ jobs:
 
     - name: Install uv
       uses: astral-sh/setup-uv@v9.0.0
+      with:
+        cache-suffix: ${{ github.workflow }}
 
     - name: Test building documentation
       run: uv run python -m sphinx docs/ docs/_build/ -b html -W
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install uv
       uses: astral-sh/setup-uv@v9.0.0
       with:
-        cache-suffix: ${{ github.workflow }}
+        save-cache: false
 
     - name: Test building documentation
       run: uv run python -m sphinx docs/ docs/_build/ -b html -W
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -21,6 +21,8 @@ jobs:
 
     - name: Install uv
       uses: astral-sh/setup-uv@v9.0.0
+      with:
+        cache-suffix: ${{ github.workflow }}
 
     - name: Install pre-commit hooks
       run: uvx pre-commit install --install-hooks
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Install uv
       uses: astral-sh/setup-uv@v9.0.0
       with:
-        cache-suffix: ${{ github.workflow }}
+        save-cache: false
 
     - name: Install pre-commit hooks
       run: uvx pre-commit install --install-hooks
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,6 +26,8 @@ jobs:
     
     - name: Install uv
       uses: astral-sh/setup-uv@v9.0.0
+      with:
+        cache-suffix: ${{ github.workflow }}
 
     # PyPI package
     - name: Build Python package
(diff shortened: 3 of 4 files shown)


hagenw:
If I understand it correctly the conflict only happens for updating the cache (re-uploading it at the end). All jobs should still be able to load the same cache, or are all the loosing jobs start with an empty cache?

In the proposed solution here we would store several caches even if they are more or less identical. Not sure if this is better.

ChristianGeng:
```
Correct — the conflict is save-time only. Losing jobs restore the shared cache
normally on later runs; nobody starts empty. And the warning only fires when the
cache key rotates (new dependency hash / first run with a new key): across the
last six pushes to `main` it appeared in 2 of 108 jobs, both on the #206 merge
that had just rotated the key — not on every push.

The storage point stands too: per-workflow suffixes would keep several
near-identical caches and give up cross-workflow sharing.

Pivoted in ef2738d: Documentation, Linter and Publish now set `save-cache:
false` (input available in setup-uv v9.0.0), and Test — the job with the fullest
environment — remains the sole writer. One writer means no reservation race and
no warning, while the single shared cache is kept.

The only cost: after a key rotation (e.g. a `pyproject.toml` change), the
non-saving workflows miss the cache once, until the next Test run saves the new
key — roughly a second or two per job for one run.
```

This writing above is too complicated.  Now more bluntly:

The behavior now is that  the setting is `save-cache: false`  in 
- Documentation, 
-  Linter
- Publish

`Test` keeps the default and is the only writer. This also eliminates the race, as 
now the writer is fixed **by designation**, not by finishing order.
always Test writes it  - because it builds the fullest environment. With only one writer there is also nothing to race.

hagenw:
Thanks, the current approach makes sense.

In the single test cache, I assume the package versions for different Python versions are anyway stored at different locations and do not override each other?

ChristianGeng:
> Thanks, the current approach makes sense.
> 
> In the single test cache, I assume the package versions for different Python versions are anyway stored at different locations and do not override each other?

Yes, they can't override each other, because the Python version is part of the cache key. There isn't one Test cache there's one per matrix entry:

-  `…ubuntu-24.04-3.10.20-5899ac3c…`
-  `…ubuntu-24.04-3.11.15-5899ac3c…`

and so on.

hagenw (approved):

